PR #35, opened 2026-04-05
Openclaw/feat branch consolidation

EricWehrly:


client/test/setup.ts
```diff
@@ -3,10 +3,14 @@
 
 import { beforeEach } from 'vitest'
 import { installMockWorker, resetMockWorkerMessageHandlers } from './utils/mock-worker'
+import { installNetworkIsolation, resetNetworkIsolation } from './utils/network-isolation'
 
 installMockWorker()
+installNetworkIsolation()
```
EricWehrly: I don't actually like having a second, stripped-down and potentially feature-differentiating file for our live tests. What if our regular setup.ts forked based on an opt-in environment type setting, and the live test configured itself for a "live-test" environment or ...something?
As long as it's opt-in to overcome the network isolation, that satisfies all the use cases I'm tracking. Open to your thoughts tho

docs/roadmaps/tech-debt.md
```diff
@@ -2,26 +2,55 @@
...
+**Context**: `LegacyStorePropsHandler` exists as a CPU-fallback path when the GPU instancing check fails. Questions to answer: (1) Does it still function at all without GPU instancing? (2) Can we test it in Docker with SwiftShader or a software renderer? (3) Is there DRY opportunity with `GpuStorePropsEventHandler`, or should they stay fully separate? (4) Is anyone likely to hit this path in practice?  
+**Source**: Apr 2026
+
+### Convention: Remove noisy JSDoc that just restates the signature
+**Priority**: Low (ongoing hygiene — apply as you touch files)  
+**Effort**: Minutes per file  
+**Context**: Remove `/** ... */` blocks that say nothing beyond what the signature already communicates. TypeScript types describe parameters better. Keep comments that explain *why*, edge cases, or non-obvious behavior. This is a "clean as you go" rule, not a separate task — check each file you're already editing.  
+**Source**: Apr 2026
+
+### Performance: Offload carpet texture generation to worker (`carpet-worker-offload`)
```
EricWehrly, line 21: You're mixing architecture, convention, deferred work, _and_ definition of our ongoing tech debt items all in one file. That's too many mixed concerns. It's all "tech debt" but we're talking different kinds of tech debt, and when we want to be mindful of 'your' context, that's something important to keep in mind.
Let's at least straighten out the most not-related concerns?

client/test/setup.ts
EricWehrly: Done — reverted to a single setup.ts. The live config (itest.live.config.ts) now sets env: { VITEST_LIVE: 'true' }, and setup.ts reads import.meta.env.VITEST_LIVE to opt out of network isolation. setup.live.ts deleted. One file, explicit opt-in.

docs/roadmaps/tech-debt.md
EricWehrly: Done — split the concerns. 	ech-debt.md is now strictly for items with deliberate work attached (TD tag IDs, architecture debt, specific fixable items). 'Clean as you go' conventions (JSDoc, file size, naming) moved to docs/guidelines/code-conventions.md. Tech-debt header updated to make the scope explicit. Phase-deferred items stay in docs/plans/ where they already were.

PR #36, opened 2026-04-05
docs: add 6.7 layout + 6.8 theme milestones from stash; archive servi…

EricWehrly:
…ce-locator spike


client/src/utils/MeshPrewarmer.ts
```diff
@@ -25,30 +26,36 @@ import { Logger } from './Logger'
...
     /**
      * Register a mesh for shader prewarm. Safe to call multiple times — each
      * call resets the debounce window so that a burst of registrations is
      * batched into a single compileAsync() call.
      *
      * No-op if prewarm has already completed or KHR is unavailable.
      */
-    public register(mesh: THREE.InstancedMesh): void {
+    public static register(mesh: THREE.InstancedMesh): void {
+        MeshPrewarmer.#inst.registerImpl(mesh)
```
EricWehrly: Could we add a class-level private getter to make this a little cleaner / easier to read? Instead of repeating our class name here and then doing `.#inst`, could we just have a private class instance `Instance` variable that resolves ` MeshPrewarmer.#inst`?

client/src/utils/MeshPrewarmer.ts
```diff
@@ -25,30 +26,36 @@ import { Logger } from './Logger'
 const DEBOUNCE_MS = 500
 
 export class MeshPrewarmer {
-    private static instance: MeshPrewarmer | null = null
+    // ES2022 private static field — truly private, not just TS-private
+    static #instance: MeshPrewarmer | null = null
     private static readonly logger = Logger.createLogFunctions(MeshPrewarmer.name)
 
+    // Lazy accessor — creates the instance on first use
+    static get #inst(): MeshPrewarmer {
```
EricWehrly: Do we have options alternative to abbreviation to differentiate?
If we're gonna do this, maybe #inst or something is the private variable, but #instance is the name that would be used elsewhere. I'd also be open to like #_instance and `instance` vs `Instance`

client/src/utils/MeshPrewarmer.ts
```diff
@@ -99,11 +106,10 @@ export class MeshPrewarmer {
     }
 
     private cleanup(): void {
-        // Remove all meshes from the temporary scene (they belong to their own managers)
         for (const child of [...this.prewarmScene.children]) {
             this.prewarmScene.remove(child)
         }
         this.completed = true
-        MeshPrewarmer.instance = null  // allow GC
+        MeshPrewarmer.#instance = null  // release for GC — one-shot lifecycle
```
EricWehrly, line 111: This is great here. Other singleton classes may have `dispose` methods that we'll want to scrutinize in the future. Probably a lot of code we _wouldn't_ want running for "singleton"s, you know? (again, all good in this one-shot, just a note for future use of the pattern)

docs/plans/service-locator-abandoned-spike.md
```diff
@@ -0,0 +1,41 @@
...
+    preference: 10
+})
+const renderer = serviceLocator.get(IStorePropsRenderer, { args: [...] })
+```
+
+---
+
+## Why it was abandoned
+
+The DI container we already have (`ServiceContainer` / `ServiceRegistration`) plus the existing capability check at registration time covers the same need more simply. Adding a full service locator layer introduced indirection without proportionate benefit.
```
EricWehrly: Now that I have this great explanation, I think we don't even keep this doc. Drop the whole thing. Good work.

client/src/utils/MeshPrewarmer.ts
EricWehrly: Resolved by dropping the getter — no abbreviation needed. #instance is the field, nullish assignment inline handles lazy init. Only one name in play.

client/src/utils/MeshPrewarmer.ts
EricWehrly: Dropped the getter entirely. Now using (MeshPrewarmer.#instance ??= new MeshPrewarmer()).registerImpl(mesh) — creates on first call, no naming needed, reads clearly.

client/src/utils/MeshPrewarmer.ts
EricWehrly: Noted — good callout for when we apply this to classes with more complex disposal logic. MeshPrewarmer is clean here because disposal IS the lifecycle end. Other singletons with teardown hooks will need their static dispose() carefully reviewed before we can fully collapse the public instance API.

docs/plans/service-locator-abandoned-spike.md
EricWehrly: Done — doc deleted.

docs/plans/phase1-remaining-work.md
```diff
@@ -0,0 +1,112 @@
...
+- ✅ MDF veneer shelf materials + brand accent colors
+- ✅ Procedural textures: wood, carpet, ceiling (off-thread workers — carpet pending)
+- ✅ GameSpotlight debug utility
+- ✅ StartupProgressUI + 5-phase startup tracking
+- ✅ FrameBudgetScheduler (stagger/cooldown now correct)
+- ✅ Shader prewarm (MeshPrewarmer)
+
+**Plumbed but unverified / needs QA:**
+- 🔄 Games appearing on shelves — event chain exists, visual confirmation pending (6.2.2.1 audit in progress)
+- 🔄 Room resizes dynamically to fit game library — implemented, not visually confirmed at scale
```
EricWehrly: this works with our current box-shaped room. We'll need to retest this when we add alternate layouts, and again in phase 2 when we add room differentiation

docs/plans/phase1-remaining-work.md
```diff
@@ -0,0 +1,112 @@
...
+- [ ] Games appear at correct positions relative to shelf geometry
+- [ ] No obvious overlaps or clipping through shelf boards
+- [ ] At least 18 games/shelf (3 per surface × 6 surfaces) renders correctly
+
+**Acceptance**: Walk through the store, games visible on shelves, artwork loading on them.
+
+### 2. Level layout — shelf/room spatial design (6.1 + 6.2.1) 🔴 IMMEDIATE
+*The store needs to feel like a store, not a corridor of floating shelves.*
+
+- [ ] Confirm alternating shelf rotation (15° offset) is implemented or implement it
```
EricWehrly: I think the scene is not doing this right now. Let's get it implemented.

docs/plans/phase1-remaining-work.md
```diff
@@ -0,0 +1,112 @@
...
+- [ ] Verify shelf spacing makes it navigable (can walk between them)
+- [ ] Room size scales reasonably for typical library (100–800 games)
+- [ ] Entrance/spawn position feels correct
```
EricWehrly: this is fine but needs polishing late in phase 2

docs/plans/phase1-remaining-work.md
```diff
@@ -0,0 +1,112 @@
...
+
+**Acceptance**: Standing at spawn, the scene reads as a store.
+
+### 3. MDF veneer shelf visual pass (6.1.1) 🟡 NEXT
+- [ ] Shelves display MDF veneer material (implemented — confirm visually)
+- [ ] Brand accent color (blue) on vertical supports
+- [ ] Shelf interior surfaces glossy white
+
+### 4. Wall/ceiling texture polish (Phase 2 capstone, but pull forward if easy)
```
EricWehrly: Why are we defining phase 2 work here?
We can set up a workflow where you try to subagent/piecemeal/break down a task ahead of our intended working target (in this case in phase 2). That's good work. But the phase 1 roadmap is the one place I'd not want that to be. 
This isn't "what we need to get done to move on from phase 1", it's "what we could be working on on the side" and we should already have a few threads right now, without pulling others...

docs/plans/phase1-remaining-work.md
```diff
@@ -0,0 +1,112 @@
...
+*The store needs to feel like a store, not a corridor of floating shelves.*
+
+- [ ] Confirm alternating shelf rotation (15° offset) is implemented or implement it
+- [ ] Verify shelf spacing makes it navigable (can walk between them)
+- [ ] Room size scales reasonably for typical library (100–800 games)
+- [ ] Entrance/spawn position feels correct
+
+**Acceptance**: Standing at spawn, the scene reads as a store.
+
+### 3. MDF veneer shelf visual pass (6.1.1) 🟡 NEXT
```
EricWehrly: I think this is fine but would like to revisit and review in later and lower priority phase 2. It looks convincing for now.

docs/plans/phase1-remaining-work.md
```diff
@@ -0,0 +1,112 @@
...
+- [ ] Shelves display MDF veneer material (implemented — confirm visually)
+- [ ] Brand accent color (blue) on vertical supports
+- [ ] Shelf interior surfaces glossy white
+
+### 4. Wall/ceiling texture polish (Phase 2 capstone, but pull forward if easy)
+- [ ] Wall planks read as planks at VR scale — may just need repeat/scale tuning
+- [ ] Ceiling popcorn visible at natural overhead view angle
+
+### 5. Game search / omnibar (6.6 UI, post-normalization)
```
EricWehrly: Note that this should tie into / DRY out with the search in the game binder...

docs/plans/phase1-remaining-work.md
```diff
@@ -0,0 +1,112 @@
...
+6. No obvious visual bugs that would embarrass you showing it
+
+That's it. Not perfect — just demo-ready for yourself.
+
+---
+
+## Parking lot (not Phase 1)
+
+- Multiple rooms
+- Categorization-based shelving (6.5)
```
EricWehrly: No I do want this as our last real feature in phase 1.

docs/plans/phase1-remaining-work.md
```diff
@@ -0,0 +1,112 @@
...
+- Layout selector (6.7)
```
EricWehrly: I do want this as part of phase 1, it just has the UI refactor work intended in front of it

When we get to building this, I'd also like us to do two research forks:
- Are there other UI elements, like the carpet editor, that we want to schedule to build?
- Are there existing UI menus, like the binder, that didn't get addressed during the UI normalization? Let's make sure we roadmap those, though they don't need to be done for us to consider phase 1 complete
- and I have a note/idea I want to do with the lighting panel, remind me if we don't fill it in somewhere by the time we get to this...

client/src/utils/MeshPrewarmer.ts
```diff
@@ -34,21 +37,20 @@ export class MeshPrewarmer {
...
      * call resets the debounce window so that a burst of registrations is
      * batched into a single compileAsync() call.
      *
      * No-op if prewarm has already completed or KHR is unavailable.
      */
-    public register(mesh: THREE.InstancedMesh): void {
+    public static register(mesh: THREE.InstancedMesh): void {
+        // ??= creates the instance on first call; subsequent calls reuse it.
+        // #instance is nulled in cleanup() once prewarm completes.
+        ;(MeshPrewarmer.#instance ??= new MeshPrewarmer()).registerImpl(mesh)
```
EricWehrly: `??= new MeshPrewarmer()` is still an action/code/operation/unit.
To me, it's more important to DRY this out and declare it in one place than to be concise. If we have no other options, let's just differentiate the names of the field and the getter, but preferably without abbreviation.

docs/roadmaps/tech-debt.md
```diff
@@ -415,31 +415,35 @@ Add a reference doc or inline comments explaining when and why to adjust each.
...
 
-Benefits: no repeated `getInstance()` calls at call sites, the null check lives in one place, and the `#instance` field is truly private (not just `private` TS).
+**Reference implementation**: `client/src/utils/MeshPrewarmer.ts` (merged Apr 2026)  
+**Note on dispose()**: Before applying to classes with complex teardown (event unregistration, GPU cleanup, etc.), review what the existing `dispose()` does. Some singletons have disposal logic that shouldn't be hidden behind a static method without careful thought.
 
-**Agree on the pattern with a poster-child class before going wide.** Best first candidate: **`DataManager`** — it's referenced in many files, its `getInstance()` calls are among the noisiest, and it has no complex constructor logic. Once `DataManager` is reshaped and reviewed, apply the pattern to `FrameBudgetScheduler`, `MeshPrewarmer`, `SharedMaterialManager`, etc.
+**Next candidates** (easiest first): `TextureLoader`, `StartupEventTracker`, `LightRegistry`, then `SharedMaterialManager`, then eventually `DataManager`.
 
-**Files tagged**: (add `// TD: singleton-pattern-refactor` as files are identified)
+**Files tagged with `// TD: singleton-pattern-refactor`**:
```
EricWehrly: Slightly misleading. Our TD convention tends to hijack the block comment at the top of the file, so an exact search for `// TD: singleton-pattern-refactor` would miss those. Leaving out the `//` doesn't set up a potential booby trap for less-informed future contexts

docs/roadmaps/phase1-ready-for-me.md
```diff
@@ -660,7 +660,52 @@ Games need to spawn onto shelves before we can "optimize" the process of doing s
...
+- Simple pattern names, not complex labels
+- Dropdown or panel in settings (post UI-normalization)
+
+### Feature 6.7.2: Layout Implementations
+- **Rectangle (current)**: Columns configurable, dynamically adjustable in UI
+- **Circle**: Radius configurable, games arranged in arc/ring
+- **Concentric Circles**: Multiple rings (the "Dreamcast logo" pattern), radius per ring, games spread outward
+- Each layout defined by a simple data structure; layouts are pluggable
+
+### Feature 6.7.3: Sorting & Filtering Prerequisites
```
EricWehrly, line 681: This is kind of at the intersection of 'categorization' and 'layout' FYI. Requires both features sufficiently implemented

docs/plans/phase1-remaining-work.md
EricWehrly: Agreed. Keeping this as immediate work: alternating shelf rotation is explicitly a verify-or-implement item under Section 2.

docs/plans/phase1-remaining-work.md
EricWehrly: Makes sense. I removed wall/ceiling polish from this Phase 1 working doc so it stays focused on Phase 1 deliverables.

docs/plans/phase1-remaining-work.md
EricWehrly: You are absolutely right. I removed Phase 2 side-work from this doc and tightened it to Phase-1-only execution items.

docs/plans/phase1-remaining-work.md
EricWehrly: Aligned — removed from this Phase 1 list; we can revisit in lower-priority Phase 2 polish.

docs/plans/phase1-remaining-work.md
EricWehrly: Added: omnibar task now explicitly calls out DRYing query/filter logic with the Game Library Binder search path.

docs/plans/phase1-remaining-work.md
EricWehrly: Confirmed and reflected: game search / omnibar is now marked as the last real Phase 1 feature.

docs/plans/phase1-remaining-work.md
EricWehrly: Captured sequencing (UI normalization first, search after) and kept search in Phase 1. I also noted your research-fork prompts for when we start building this.

client/src/utils/MeshPrewarmer.ts
EricWehrly: Implemented this exactly: lazy-init now lives in one class-level private getter (`#current`) and `register()` delegates to it.

docs/roadmaps/tech-debt.md
EricWehrly: Great catch. Updated wording to avoid the literal `// TD:` search trap — it now says to search for `singleton-pattern-refactor` in file header comments.

docs/roadmaps/phase1-ready-for-me.md
EricWehrly: Updated the 6.7 prerequisite text to explicitly say it sits at the categorization/layout intersection and depends on both being sufficiently implemented.

docs/plans/phase1-remaining-work.md
EricWehrly: Good call — added an explicit retest note under 'Plumbed but unverified' to re-validate this when alternate layouts and room differentiation are introduced.

docs/plans/ui-design-tokens.md
```diff
@@ -0,0 +1,100 @@
+# UI Design Tokens — Phase 1
```
EricWehrly, line 1: This whole thing is good work, but should probably just be a literal CSS file. We can review that.

docs/plans/ui-normalization-audit.md
```diff
@@ -0,0 +1,47 @@
...
+## Overview
+This audit covers the current state of the UI in the Steam Brick and Mortar client. The UI is built using vanilla TypeScript and DOM manipulation, with styles split between external CSS files, template-embedded styles, and direct inline style manipulation in TypeScript.
+
+## Component Inventory
+
+| Component | Files | Style Method | Consistency Notes |
+|-----------|-------|--------------|-------------------|
+| **LightingControlsPanel** | `LightingControlsPanel.ts`, `lighting-controls-panel.css` | External CSS + Inline (visibility) | Uses a mix of specific classes and general selectors. Some hardcoded hex/rgba. |
+| **PerformanceMonitorUI** | `PerformanceMonitor.ts` | **Inline Only** | Highly inconsistent with the rest of the app. Hardcoded colors like `#00ff00`, `#ffff00`, `#ff0000`. |
+| **ToastManager** | `ToastManager.ts`, `toast.css` | External CSS | Relatively clean, but uses hardcoded colors in CSS. |
```
EricWehrly: Good add, but I'm not sure I want to keep the toastmanager. Let's flag this as something to leave as-is until later in phase 2. We may even want to consider just shutting it off, with our other UI stuff. But ... leave it for now, lowest priority. If we have spare cycles, great. After the other UI work.

docs/plans/ui-normalization-audit.md
```diff
@@ -0,0 +1,47 @@
...
+| **PauseMenu** | `PauseMenuManager.ts`, `PauseMenuPanel.ts`, `shared-components.css` | External CSS (Shared) | The most "normalized" part of the app, but still has hardcoded values in the shared CSS. |
```
EricWehrly, line 13: There's quite a few panels in here, more than just `GraphicsSettingsPanel`, and more than just one for cache.
We'll want to address all of the pause menu panels, as well as the menu itself overall.

docs/plans/phase1-remaining-work.md
```diff
@@ -0,0 +1,142 @@
...
+- 🔄 Games appearing on shelves — event chain exists, visual confirmation pending (6.2.2.1 audit in progress)
+- 🔄 Room resizes dynamically to fit game library — implemented, not visually confirmed at scale; retest when alternate layouts and room differentiation land
+- 🔄 Shelf capacity calculation — `GameLayoutConstants` hardcoded values may need tuning
+- 🔄 GameStart event → model generation — task 6.0.1.2 (models deferred until GameStart) may already be done
+
+---
+
+## Remaining work (ordered)
+
+### 1. Verify & fix game spawning on shelves (6.2.2.1) 🔴 IMMEDIATE
```
EricWehrly: Take this as my verification: 
Prior to any rotation implementation, `Verify & fix game spawning on shelves` is working fine. I think all of these are done and shouldn't be considered pending.

docs/plans/phase1-remaining-work.md
```diff
@@ -0,0 +1,142 @@
...
+**Acceptance**: Walk through the store, games visible on shelves, artwork loading on them.
+
+### 2. Shelf-in-room layout (6.1 + 6.2.1) 🔴 PILLAR 1
+*The store needs to feel like a store.*
+
+- [ ] Shelves are arranged in an intentional layout (rows, navigable aisles)
+- [ ] Confirm or implement alternating shelf rotation (15° offset)
+- [ ] Room size scales reasonably for typical library (100–800 games)
+- [ ] Layout is configurable (row count / spacing) even if we ship with one preset
+- [ ] Entrance/spawn position feels correct
```
EricWehrly: The entrance/spawn position is good. We'll probably want to review after we start implementing VR, but don't need to look at it sooner.

client/src/types/InteractionEvents.ts
```diff
@@ -194,6 +194,13 @@ export interface AllBatchesCompleteEvent extends BaseInteractionEvent {
     // Pure terminal signal: all batches are complete.
 }
 
+export interface GameSelectedEvent extends BaseInteractionEvent {
+    /** App ID of the selected game */
+    appid: number | string
+    /** Display name, if known */
+    name?: string
```
EricWehrly: Do we need this optional name field? Are we gonna look it up every time anyway?

client/src/ui/binder/GameLibraryBinderUI.ts
```diff
@@ -448,6 +454,29 @@ export class GameLibraryBinderUI {
...
     }
     
+    /**
+     * Open the detail panel for a specific game by appid.
+     * Called externally (e.g. from scene raycast click).
+     */
+    public openGameDetail(appid: number | string): void {
+        // Search filteredGames first (fast path if binder is open/loaded),
+        // then fall back to full library from DataManager
+        let game = this.state.filteredGames.find(g => String(g.appid) === String(appid))
```
EricWehrly, line 469: Ah, line 464 won't "work" (this.state contains no filteredGames) if the binder menu itself hasn't been opened yet.

Let's try to make it so that the state can be fetched if not initialized.

This is starting to introduce/expand business logic beyond UI logic ...

Also ... we may want to have a GameUtility method or _something_ to do global game object resolution and not need each interfacing class to figure out it's own way ... seems like a good tech debt refactor for later.

client/src/ui/binder/GameLibraryBinderUI.ts
```diff
@@ -448,6 +454,29 @@ export class GameLibraryBinderUI {
...
+        if (!game) {
+            const allGames = this.dataManager.get<SteamGameData[]>('steam.games') ?? []
+            game = allGames.find(g => String(g.appid) === String(appid))
+        }
+        if (game) {
+            this.selectGame(game)
```
EricWehrly, line 475: selectGame results in a panel being added to the binder container on line 596. So the detail panel isn't set up to be opened if the binder isn't open.
We want it to be able to float by itself. It really doesn't need to be child to the binder for any other reason than
"hide the detail panel when the binder closes", which we can call explicitly rather than relying on parentage.

Now that we're separating the concerns, it probably even makes sense to have the binder and the detail page go in separate files....

client/src/ui/binder/GameLibraryBinderUI.ts
```diff
@@ -483,10 +512,29 @@ export class GameLibraryBinderUI {
...
+        // Build categories/genres block
+        const genres = game.genres?.map(g => g.description) ?? []
+        const steamCategories = game.categories?.map(c => c.description) ?? []
+        const categoriesHtml = genres.length > 0 || steamCategories.length > 0 ? `
+            <div class="detail-categories">
+                <div class="detail-section-label">Categories</div>
+                ${genres.length > 0 ? `
+                <div class="detail-category-group">
+                    <span class="detail-category-label">Genres</span>
+                    <div class="detail-tags">${genres.map(g => `<span class="detail-tag">${this.escapeHtml(g)}</span>`).join('')}</div>
```
EricWehrly: Instead of big white text for the categories list on lines 525 and 530, could we do "slugs" (basically just like spans or whatever with a lil padding and a background-color. also foreground color should be less striking than white)
There may be a "steam slug style" we want to yoink for this rather than trying to figure that out
Either way, these need a lil spacing. Probly `join(' ')` at _least_!

client/src/ui/binder/GameLibraryBinderUI.ts
```diff
@@ -483,10 +512,29 @@ export class GameLibraryBinderUI {
...
+                </div>` : ''}
+                ${steamCategories.length > 0 ? `
+                <div class="detail-category-group">
+                    <span class="detail-category-label">Features</span>
+                    <div class="detail-tags">${steamCategories.map(c => `<span class="detail-tag">${this.escapeHtml(c)}</span>`).join('')}</div>
```
EricWehrly: If we can get / find icons for these categories (not that we assign, but that steam has defined somehow) that would be wonderful. I'd like them prepended on both/either sets of categories (Genres and Features) where they're available. Some with icons and some without is 100% tolerable, I'd like to preview.

client/src/ui/binder/GameLibraryBinderUI.ts
```diff
@@ -522,6 +570,8 @@ export class GameLibraryBinderUI {
                     </div>
                 </div>
 
+                ${categoriesHtml}
+
                 <div class="detail-artwork">
```
EricWehrly: We should have a class like "section" (or vertically-stacked? Whatever makes sense) and define our section separators using a border-top for all _except_ the first one (or if that's prohibitive for some reason, all except last with a border-bottom would be equally fine, whichever's easier in CSS -- this should replace existing separators where we define em)

docs/plans/open-subagent-threads.md
```diff
@@ -0,0 +1,64 @@
+# Open Subagent Threads
```
EricWehrly, line 1: Let's update the readme, or at least your openclaw workspace, to make sure we keep a handle on this. 
We're using this as our "as we have spare cycles" container to pull from, kind of an alternate to 'roadmap'.

Did we not have any refactoring threads we wanted to push?
I still want to do the getInstance stuff. 
The DataManager changes are significant enough I want to do those when we get a fresh branch, but it's still a thread I want to pursue as it becomes available.

docs/plans/open-subagent-threads.md
```diff
@@ -0,0 +1,64 @@
...
+- `client/src/scene/GpuStorePropsRenderer.ts` — where shelf placement happens (needs group awareness)
+- `client/src/scene/spawning/GameBoxSpawner.ts` — game→shelf assignment
+
+**Prep Notes (main session)**:
+- Confirm what % of a typical library actually has genres populated (may need a dev-mode log)
+- Decide: group by `genres[0].description` (primary genre) or something smarter?
+
+---
+
+## Completed Threads
```
EricWehrly: Let's not track 'done' threads right now. Let's rely on our existing documentation to hold that information.

client/src/ui/binder/BinderGameDetailPanel.ts
```diff
@@ -0,0 +1,155 @@
...
+            </div>
+        ` : ''
+
+        const jsonBlob = JSON.stringify(game, null, 2)
+
+        const panel = document.createElement('div')
+        panel.id = 'binder-detail-panel'
+        panel.className = 'detail-panel'
+
+        panel.innerHTML = `
```
EricWehrly: Can we still do things like `${this.escapeHtml(game.name)}` if we use html template files? If it works, I'd like to go ahead and migrate these larger html blocks to template files please.

EricWehrly: Yes — \ works fine in \.html\ template files when imported as raw strings via Vite's \?raw\ suffix (or \\\import template from './detail.html?raw'\\\). The template is just a string at that point and we do the substitution ourselves before \innerHTML\ assignment — same as now, just the markup lives in a dedicated file instead of inline. I'll migrate the BinderGameDetailPanel markup to a \.html\ template file as part of the UI normalization pass — it pairs naturally with the design token migration.

client/src/types/InteractionEvents.ts
EricWehrly: Good call - the name field is dead weight. The receiver always does a DataManager lookup by appid anyway. Dropping it.

client/src/scene/interaction/SceneClickGameBoxRaycast.ts
```diff
@@ -196,16 +195,13 @@ export class SceneClickGameBoxRaycast {
...
     private highlightHit(hit: SceneGameBoxHit): void {
-        const spotlightTarget = hit.name ?? hit.appid
-        if (spotlightTarget !== undefined) {
-            const spotlight = GameSpotlight.getInstance()
-            if (spotlight) {
-                spotlight.spotlight(spotlightTarget)
-            } else if (this.enableDebugLogs) {
-                console.log('🎯 [SceneClickGameBoxRaycast] GameSpotlight not ready yet')
-            }
+        const appid = hit.appid
```
EricWehrly, line 198: we don't need this line, can just do `if (hit.appid` below.

client/src/ui/binder/BinderGameDetailPanel.ts
```diff
@@ -0,0 +1,113 @@
...
+            .replace(/\{\{headerUrl\}\}/g, headerUrl)
+            .replace(/\{\{libraryUrl\}\}/g, libraryUrl)
+            .replace(/\{\{gameName\}\}/g, this.escapeHtml(game.name))
+            .replace(/\{\{appid\}\}/g, String(game.appid))
+            .replace(/\{\{playtimeHours\}\}/g, String(playtimeHours))
+            .replace(/\{\{playtime2WeeksBlock\}\}/g, playtime2WeeksBlock)
+            .replace(/\{\{categoriesBlock\}\}/g, categoriesBlock)
+            .replace(/\{\{jsonBlob\}\}/g, this.escapeHtml(jsonBlob))
+
+        const panel = document.createElement('div')
```
EricWehrly, line 64: Maybe we repaint the panel when updating, but only create it once? Then some of these later lines only happen at build-time, not show-time. It should result in a better UX.

client/src/ui/binder/detail-panel.html
```diff
@@ -0,0 +1,46 @@
...
+        {{playtime2WeeksBlock}}
+        <div class="detail-stat">
+            <div class="detail-stat-label">App ID</div>
+            <div class="detail-stat-value">{{appid}}</div>
+        </div>
+    </div>
+
+    {{categoriesBlock}}
+
+    <div class="detail-artwork">
```
EricWehrly: I think these sections, or vertical-stacks should have separators that are either:
- border-bottom excluding the last in the stack (no need for separator with the bottom of the UI)
or
- border-top excluding the first (because we have a header image)

whichever is easier

These borders should replace any existing dividers

docs/plans/ui-design-tokens.md
EricWehrly: Agreed - a literal CSS file is the right home for this. The .md doc was useful as a planning artifact but the tokens should live as a real :root { --var: value } file that components can import. I'll move it to client/src/ui/tokens.css (or similar) during the Phase B component work, after we've confirmed the token values look right in practice.

docs/plans/ui-normalization-audit.md
EricWehrly: Noted - leaving ToastManager as-is until later Phase 2 at earliest. May get shut off entirely when the broader UI gets tidied up.

docs/plans/ui-normalization-audit.md
EricWehrly: Audit updated to list all pause menu panels (tabs + menu container), LightingControlsPanel, CacheManagementUI, GraphicsSettingsPanel, PerformanceMonitorUI, StartupProgressUI, and the binder/detail panel as normalization targets. Not just the settings + cache panels.

docs/plans/phase1-remaining-work.md
EricWehrly: Verified and marked done in phase1-remaining-work.md. Spawn/entrance position also confirmed good - noted to re-review post-VR but not before.

docs/plans/phase1-remaining-work.md
EricWehrly: Noted and updated in phase1-remaining-work.md - entrance/spawn confirmed good, deferred re-review to after VR implementation starts.

docs/plans/open-subagent-threads.md
EricWehrly: Updated open-subagent-threads.md: removed Completed/Abandoned sections (relying on git history and roadmap docs for that), added getInstance singleton refactor thread, tightened scope descriptions to one-file-at-a-time bite sizes.

docs/plans/open-subagent-threads.md
EricWehrly: Removed the Completed Threads section from open-subagent-threads.md. Completed work lives in git history and roadmap docs.

client/src/ui/binder/BinderGameDetailPanel.ts
EricWehrly: Good idea. Tracked in open-subagent-threads.md under UI Normalization - panel can be created once and repainted on update rather than fully rebuilt on show(). Will implement during the Phase B normalization pass.

client/src/ui/binder/detail-panel.html
EricWehrly: Adding border-bottom separators between content sections (excluding the last). Will replace any existing dividers.

PR #37, opened 2026-04-06
Openclaw/feat 6.5 categories

EricWehrly:


client/src/scene/batch/BatchCoordinator.ts
```diff
@@ -218,6 +218,7 @@ export class BatchCoordinator<T> {
                 games: batchData.games,
                 batchIndex,
                 totalBatches,
+                shelfLabel: batchData.shelfLabel,
```
EricWehrly, line 221: Batches and shelves shouldn't have anything to do with one another. Batches are just as things things arrive, to control flow and rate when hitting providing resources. Shelves are where games land based on information that is entirely subsequent to this...

Anyway, I don't understand why batchData would have shelfLabel, let alone pass it here.

PR #38, opened 2026-04-06
feat(categories-v2): clean category/sign pipeline + signage regressio…

EricWehrly:
…n tests

Clean branch built from master with only intended category-related changes:

- Add CategoryAssigner with KNOWN_GENRES canonical list + case-insensitive lookup
- Add CategoryAssigner tests (9 unit tests)
- Add CategorySignSystem and ShelfSectionPlanner (single-pass signs after all batches)
- Remove batch-time category inference from GameBoxSpawner
- GpuStorePropsRenderer accumulates games safely across init/ordering
- Fix sign anchor collisions via minimum-advance anchor policy (small groups no longer stack to one shelf)
- Fix SignageRenderer shared-canvas bug (all signs showing same text)
- Add SignageRenderer regression tests (independent canvas per sign)
- Add CategoryReferencePanel UI (+ CSS) and wire to SteamBrickAndMortarApp
- Update InteractionEvents.ShelfCreatedEvent optional fields used by GPU pipeline
- Carry roadmap/tech-debt/open-thread docs updates

### DIFF
--- a/client/src/core/SteamBrickAndMortarApp.ts
+++ b/client/src/core/SteamBrickAndMortarApp.ts
@@ -14,6 +14,7 @@
 
 import * as THREE from 'three'
 import { PerformanceMonitorUI, type PerformanceStats, ToastManager, UIManager, StartupProgressUI, GameLibraryBinderUI } from '../ui'
+import { CategoryReferencePanel } from '../ui/CategoryReferencePanel'
 import { SteamUICoordinator, WebXRUICoordinator, SystemUICoordinator } from '../ui/coordinators'
 import { SceneManager, SceneCoordinator } from '../scene'
 import { DebugStatsProvider } from './DebugStatsProvider'
@@ -64,6 +65,7 @@ export class SteamBrickAndMortarApp {
     private appSettings: AppSettings
     private compassRose?: CompassRose
     private gameLibraryBinder?: GameLibraryBinderUI
+    private categoryReferencePanel?: CategoryReferencePanel
     
     // Startup tracking
     private startupTracker: StartupEventTracker
@@ -273,6 +275,10 @@ export class SteamBrickAndMortarApp {
             this.gameLibraryBinder = GameLibraryBinderUI.getInstance()
             this.gameLibraryBinder.init()
             this.startupTracker.logEvent(StartupPhase.PrewarmEncore, 'Game Library Binder UI initialized')
+
+            this.categoryReferencePanel = new CategoryReferencePanel()
+            this.categoryReferencePanel.init()
+            this.startupTracker.logEvent(StartupPhase.PrewarmEncore, 'Category Reference Panel initialized')
 
             // Initialize system UI coordinator (lighting panel, debug panels, etc.)
             await this.systemUICoordinator.init(this.sceneManager.getRenderer())
--- a/client/src/core/SteamBrickAndMortarApp.ts
+++ b/client/src/core/SteamBrickAndMortarApp.ts
@@ -14,7 +14,6 @@
 
 import * as THREE from 'three'
 import { PerformanceMonitorUI, type PerformanceStats, ToastManager, UIManager, StartupProgressUI, GameLibraryBinderUI } from '../ui'
-import { CategoryReferencePanel } from '../ui/CategoryReferencePanel'
 import { SteamUICoordinator, WebXRUICoordinator, SystemUICoordinator } from '../ui/coordinators'
 import { SceneManager, SceneCoordinator } from '../scene'
 import { DebugStatsProvider } from './DebugStatsProvider'
@@ -65,7 +64,6 @@ export class SteamBrickAndMortarApp {
     private appSettings: AppSettings
     private compassRose?: CompassRose
     private gameLibraryBinder?: GameLibraryBinderUI
-    private categoryReferencePanel?: CategoryReferencePanel
     
     // Startup tracking
     private startupTracker: StartupEventTracker
@@ -275,10 +273,6 @@ export class SteamBrickAndMortarApp {
             this.gameLibraryBinder = GameLibraryBinderUI.getInstance()
             this.gameLibraryBinder.init()
             this.startupTracker.logEvent(StartupPhase.PrewarmEncore, 'Game Library Binder UI initialized')
-
-            this.categoryReferencePanel = new CategoryReferencePanel()
-            this.categoryReferencePanel.init()
-            this.startupTracker.logEvent(StartupPhase.PrewarmEncore, 'Category Reference Panel initialized')
 
             // Initialize system UI coordinator (lighting panel, debug panels, etc.)
             await this.systemUICoordinator.init(this.sceneManager.getRenderer())
--- a/client/src/scene/CategorySignSystem.ts
+++ b/client/src/scene/CategorySignSystem.ts
@@ -1,0 +1,164 @@
+/**
+ * CategorySignSystem
+ *
+ * Generic scene-space sign system for category labels.
+ *
+ * A "sign" is a canvas-texture plane mesh with configurable:
+ * - Text and color
+ * - Mount style: above-shelf, wall-mounted, ceiling-hung
+ *
+ * This sits on top of SignageRenderer's createSign() primitive.
+ * Visual styles (Blockbuster, neon, flat, etc.) are applied by passing
+ * a SignStyle config. Multiple realizations share one code path.
+ *
+ * Phase 1: above-shelf signs only.
+ * Phase 2: wall + ceiling mount styles.
+ */
+
+import * as THREE from 'three'
+import { SignageRenderer, type SignageConfig } from './SignageRenderer'
+
+// ─── Style definitions ────────────────────────────────────────────────────────
+
+export interface SignStyle {
+    backgroundColor: number
+    textColor: number
+    width: number
+    height: number
+}
+
+export const SignStyles = {
+    /** Default Blockbuster-style blue sign */
+    Category: {
+        backgroundColor: 0x1a3a5c,
+        textColor: 0xffffff,
+        width: 2.2,
+        height: 0.38,
+    } satisfies SignStyle,
+
+    /** Accent color — for featured/highlighted sections */
+    Featured: {
+        backgroundColor: 0x8b0000,
+        textColor: 0xffffff,
+        width: 2.2,
+        height: 0.38,
+    } satisfies SignStyle,
+} as const
+
+// ─── Mount styles ─────────────────────────────────────────────────────────────
+
+export type SignMountStyle = 'above-shelf' | 'wall' | 'ceiling'
+
+export interface SignMount {
+    style: SignMountStyle
+    /**
+     * Y offset from the mount anchor point.
+     * For above-shelf: offset above the shelf top (default 0.3m).
+     * For ceiling: offset below ceiling plane.
+     * For wall: offset from wall surface.
+     */
+    yOffset?: number
+}
+
+// ─── Category sign descriptor ─────────────────────────────────────────────────
+
+export interface CategorySignDescriptor {
+    label: string
+    anchorPosition: THREE.Vector3   // World position to anchor sign to (e.g. shelf center)
+    mount: SignMount
+    style?: SignStyle
+}
+
+// ─── System ───────────────────────────────────────────────────────────────────
+
+export class CategorySignSystem {
+    private readonly renderer: SignageRenderer
+    private readonly scene: THREE.Scene
+    private readonly signs: Map<string, THREE.Mesh> = new Map()
+
+    private static readonly ABOVE_SHELF_DEFAULT_Y_OFFSET = 0.6
+    private static readonly SIGN_Z_FACE_PLAYER = 0.01 // slight forward push to avoid z-fighting
+
+    constructor(scene: THREE.Scene) {
+        this.scene = scene
+        this.renderer = new SignageRenderer()
+    }
+
+    /**
+     * Create or update a category sign.
+     * If a sign with this label already exists, it is replaced.
+     */
+    public setSign(descriptor: CategorySignDescriptor): THREE.Mesh {
+        // Remove existing sign with same label if present
+        this.removeSign(descriptor.label)
+
+        const style = descriptor.style ?? SignStyles.Category
+        const signPos = this.resolvePosition(descriptor.anchorPosition, descriptor.mount)
+
+        const config: SignageConfig = {
+            text: descriptor.label,
+            position: signPos,
+            backgroundColor: style.backgroundColor,
+            textColor: style.textColor,
+            width: style.width,
+            height: style.height,
+        }
+
+        const mesh = this.renderer.createSign(config)
+        mesh.userData.categoryLabel = descriptor.label
+        mesh.userData.mountStyle = descriptor.mount.style
+
+        this.scene.add(mesh)
+        this.signs.set(descriptor.label, mesh)
+
+        return mesh
+    }
+
+    /** Remove a named sign from the scene. */
+    public removeSign(label: string): void {
+        const existing = this.signs.get(label)
+        if (existing) {
+            this.scene.remove(existing)
+            const mat = existing.material as THREE.MeshStandardMaterial
+            mat.map?.dispose()
+            mat.dispose()
+            existing.geometry.dispose()
+            this.signs.delete(label)
+        }
+    }
+
+    /** Remove all signs. */
+    public clearAll(): void {
+        for (const label of [...this.signs.keys()]) {
+            this.removeSign(label)
+        }
+    }
+
+    public dispose(): void {
+        this.clearAll()
+        this.renderer.dispose()
+    }
+
+    // ─── Position resolution ──────────────────────────────────────────────────
+
+    private resolvePosition(anchor: THREE.Vector3, mount: SignMount): THREE.Vector3 {
+        switch (mount.style) {
+            case 'above-shelf': {
+                const yOff = mount.yOffset ?? CategorySignSystem.ABOVE_SHELF_DEFAULT_Y_OFFSET
+                return new THREE.Vector3(
+                    anchor.x,
+                    anchor.y + yOff,
+                    anchor.z - CategorySignSystem.SIGN_Z_FACE_PLAYER // face player
+                )
+            }
+            case 'wall':
+            case 'ceiling': {
+                // TODO: Phase 2 — wall and ceiling mount resolution
+                // Wall: position at anchor.x, anchor.y + yOffset, against nearest wall Z
+                // Ceiling: position at anchor.x, ceilingHeight - yOffset, anchor.z
+                const yOff = mount.yOffset ?? 0
+                return new THREE.Vector3(anchor.x, anchor.y + yOff, anchor.z)
+            }
+        }
+    }
+}
--- a/client/src/scene/CategorySignSystem.ts
+++ b/client/src/scene/CategorySignSystem.ts
@@ -2,6 +2,7 @@
  * CategorySignSystem
  *
  * Generic scene-space sign system for category labels.
+ * Tech debt link: docs/roadmaps/tech-debt.md → "Category System Tech Debt / CategorySignSystem → SceneSignManager rename"
  *
  * A "sign" is a canvas-texture plane mesh with configurable:
  * - Text and color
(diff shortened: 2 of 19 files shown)


client/src/core/SteamBrickAndMortarApp.ts
```diff
@@ -274,6 +276,10 @@ export class SteamBrickAndMortarApp {
             this.gameLibraryBinder.init()
             this.startupTracker.logEvent(StartupPhase.PrewarmEncore, 'Game Library Binder UI initialized')
 
+            this.categoryReferencePanel = new CategoryReferencePanel()
```
EricWehrly: nah put this guy in the uicoordinator
maybe make yourself a little memory note not to just spam stuff in here and instead put the UI with the UI, etc.

client/src/scene/categorization/CategoryAssigner.ts
```diff
@@ -0,0 +1,95 @@
...
+    label: string
+    games: SteamGameData[]
+}
+
+/**
+ * Hardcoded list of recognised genre names, in display-canonical casing.
+ * Case-insensitive lookup so "Free to Play" / "Free To Play" both match.
+ * Unrecognised genres fall into "Other".
+ */
+export const KNOWN_GENRES: ReadonlyArray<string> = [
```
EricWehrly, line 27: this was meant to be a hack but maybe we keep it until we refactor for the different types and sources of categories and have a more finalized view of how we want to do this...

client/src/scene/categorization/CategoryAssigner.ts
```diff
@@ -0,0 +1,95 @@
...
+const GENRE_LOOKUP: ReadonlyMap<string, string> = new Map(KNOWN_GENRES.map(g => [g.toLowerCase(), g]))
+
+/**
+ * Sort comparator for genre-first, playtime-second ordering.
+ * Games with recognised genres are grouped together; within each genre,
+ * sorted by playtime descending. Games with no recognised genre sort last.
+ *
+ * Safe to use as the sortFn option in SteamApiClient.loadGamesProgressively.
+ */
+export function genrePlaytimeSortFn(
```
EricWehrly: I would prefer we have a generic "sort by playtime" function. If typescript is really our friend here, we can write a generic "sort by field" function, that takes the field in, and then "sort by play time" just becomes a call to "sort by field" with "play time" as a parameter ....

client/src/scene/categorization/CategoryAssigner.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        const groupsMap = new Map<string, SteamGameData[]>()
+        const otherGroupName = 'Other'
+        let noGenreCount = 0
+
+        for (const game of games) {
+            let genre: string
+            if (game.genres && game.genres.length > 0) {
+                const raw = game.genres[0].description
+                const canonical = GENRE_LOOKUP.get(raw.toLowerCase())
```
EricWehrly, line 83: I may be thinking of features from another language, but I thought with certain kinds of collections, we could run a LINQ-like expression (What's the generic term for that?) for like `GENRE_LOOKUP.get(game?.genres?[0]?.description?.toLowerCase()` (admittedly that coalescing may be too much)

client/src/scene/categorization/CategoryAssigner.ts
```diff
@@ -0,0 +1,95 @@
...
+                if (canonical !== undefined) {
+                    genre = canonical
+                } else {
+                    CategoryAssigner.logger.warn(`[CAT-DEBUG] Unrecognized genre "${raw}" for "${game.name}" (appid ${game.appid}) ? Other`)
+                    genre = otherGroupName
+                    noGenreCount++
+                }
+            } else {
+                genre = otherGroupName
```
EricWehrly, line 90: You repeat your else blocks here. Seems unnecessary.
Even more, we could either initialize genre with the otherGroupName value, or maybe advantage a ternary `?`?

client/src/scene/categorization/CategoryAssigner.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        if (noGenreCount > 0) {
+            CategoryAssigner.logger.warn(`[CAT-DEBUG] ${noGenreCount}/${games.length} games had no recognised genre � landed in "Other".`)
+        }
+
+        const shelfGroups: ShelfGroup[] = Array.from(groupsMap.entries()).map(([genre, groupGames]) => ({
+            genre, label: genre, games: groupGames
+        }))
+
+        shelfGroups.sort((a, b) => {
```
EricWehrly, line 109: Are we sorting a collection that's gonna get sorted anyway?

client/src/scene/spawning/GameBoxSpawner.ts
```diff
@@ -15,154 +15,105 @@ import { Logger } from '../../utils/Logger'
...
             if (gameIndex >= games.length) break
-            
             const frontGames = games.slice(gameIndex, gameIndex + GameLayoutConstants.GAMES_PER_SURFACE)
             if (frontGames.length > 0) {
                 this.createGameBoxes(shelfPosition, surface, frontGames, ShelfSide.Front, shelfRotationY)
                 gameIndex += frontGames.length
             }
-            
             if (gameIndex < games.length) {
-                // TD: wall-shelf-back-side — wall-mounted shelves should not fill the back side.
```
EricWehrly, line 163: do not delete `// TD` lines. You should know what those are.
It'll make sense to delete them when we address them, but not ... for whatever reason you may have put in this commit message.
Write yourself a rule to prevent doing this again

client/src/scene/ShelfSectionPlanner.ts
```diff
@@ -0,0 +1,134 @@
...
+            signYOffset: config.signYOffset ?? DEFAULT_SIGN_Y_OFFSET,
+            signMountStyle: config.signMountStyle ?? DEFAULT_MOUNT_STYLE,
+        }
+    }
+
+    /**
+     * Accumulate games from one batch.
+     * Does NOT place any signs � signs are deferred to planSections().
+     */
+    public onBatchPlaced(
```
EricWehrly: what the sam fuck is ` public onBatchPlaced(`
no
private
subscribe to event in class
don't pass events through classes
that's dumb
stop it

client/src/scene/CategorySignSystem.ts
```diff
@@ -0,0 +1,164 @@
...
+        this.renderer = new SignageRenderer()
+    }
+
+    /**
+     * Create or update a category sign.
+     * If a sign with this label already exists, it is replaced.
+     */
+    public setSign(descriptor: CategorySignDescriptor): THREE.Mesh {
+        // Remove existing sign with same label if present
+        this.removeSign(descriptor.label)
```
EricWehrly, line 97: We shouldn't do update this way at all, but I don't think we're really updating right now, so we can leave this as tech debt for later

client/src/scene/CategorySignSystem.ts
```diff
@@ -0,0 +1,164 @@
...
+    private readonly renderer: SignageRenderer
+    private readonly scene: THREE.Scene
+    private readonly signs: Map<string, THREE.Mesh> = new Map()
+
+    private static readonly ABOVE_SHELF_DEFAULT_Y_OFFSET = 0.6
+    private static readonly SIGN_Z_FACE_PLAYER = 0.01 // slight forward push to avoid z-fighting
+
+    constructor(scene: THREE.Scene) {
+        this.scene = scene
+        this.renderer = new SignageRenderer()
```
EricWehrly, line 88: Do we need the signagerenderer to be an instance, or can we expose singleton methods?

client/src/scene/CategorySignSystem.ts
```diff
@@ -0,0 +1,164 @@
...
+export class CategorySignSystem {
+    private readonly renderer: SignageRenderer
+    private readonly scene: THREE.Scene
+    private readonly signs: Map<string, THREE.Mesh> = new Map()
+
+    private static readonly ABOVE_SHELF_DEFAULT_Y_OFFSET = 0.6
+    private static readonly SIGN_Z_FACE_PLAYER = 0.01 // slight forward push to avoid z-fighting
+
+    constructor(scene: THREE.Scene) {
+        this.scene = scene
```
EricWehrly, line 87: we can get scene from the datamanager, and should generally defer to doing so, so that we only have one access pattern and can work to replace it if desired
In fact, I'm hoping you pick this up and can set a todo for us to build a scenemanager with a static get for the scene, that pulls from the datamanager, and allows scenes to have an index enum. Then each time we create a scene, we can register it with the index.

client/src/scene/GpuStorePropsRenderer.ts
```diff
@@ -162,21 +173,24 @@ export class GpuStorePropsRenderer implements IStorePropsRenderer {
...
     }
-    
+
     private async handleInitialBatch(event: CustomEvent<BatchReadyForPlacementEvent>): Promise<void> {
-        const { totalBatches } = event.detail
-        
-        // Only handle the first batch for initialization
+        const { totalBatches, batchIndex, games } = event.detail
+
+        const primaryGenre = getPrimaryGenreFromBatch(games as readonly SteamGameData[])
```
EricWehrly, line 183: Thinking about it, the event should emit readonly everything all the time, right?
....that's probably another piece of tech debt we'll need to pursue...

client/src/scene/GpuStorePropsRenderer.ts
```diff
@@ -202,70 +216,75 @@ export class GpuStorePropsRenderer implements IStorePropsRenderer {
...
         // Fast path: Already ready
         if (this.isShelfRendererReady || this.instancedShelfRenderer.isReady()) {
             this.isShelfRendererReady = true
             return
         }
-        
+
         // Wait for RendererReady event
-        console.log('⏳ Waiting for RendererReady event...')
+        console.log('? Waiting for RendererReady event...')
```
EricWehrly: why drop emoji?

client/src/scene/GpuStorePropsRenderer.ts
```diff
@@ -202,70 +216,75 @@ export class GpuStorePropsRenderer implements IStorePropsRenderer {
...
+
         this.calculateShelfBoundsAndLayout(totalShelves)
     }
-    
+
     private calculateShelfRowZ(row: number): number {
         const normalRowZ = VRLayoutUtils.calculateOptimalRowPosition(row)
 
-        // Odd rows are rotated 180° and should sit closer to previous even row
+        // Odd rows are rotated 180� and should sit closer to previous even row
```
EricWehrly, line 287: replaced degrees symbol with unknown symbol

client/src/scene/GpuStorePropsRenderer.ts
```diff
@@ -297,57 +316,69 @@ export class GpuStorePropsRenderer implements IStorePropsRenderer {
...
     private ensureShelfPositionAllocated(batchIndex: number): void {
         if (batchIndex < this.shelfPositions.length) {
             return
         }
-        
+
         const oldLength = this.shelfPositions.length
         const progress = this.batchCoordinator.getProgress()
-        console.warn(`⚠️ BATCH COUNT MISMATCH: Received batch ${batchIndex + 1} but only allocated ${oldLength} positions`)
+        console.warn(`?? BATCH COUNT MISMATCH: Received batch ${batchIndex + 1} but only allocated ${oldLength} positions`)
```
EricWehrly: several of these dropped emojis, please check the whole file
whoever formatted this sucked

client/src/scene/GpuStorePropsRenderer.ts
```diff
@@ -503,22 +532,23 @@ export class GpuStorePropsRenderer implements IStorePropsRenderer {
...
-        
+
         // Dispose game box renderer and its GPU resources
         this.gameBoxRenderer?.dispose()
         this.gameBoxRenderer = null
-        
+
         this.instancedShelfRenderer?.dispose()
-        
+        this.shelfSectionPlanner.dispose()
```
EricWehrly, line 542: why otherwise question mark but not here?

client/src/scene/CategorySignSystem.ts
```diff
@@ -0,0 +1,164 @@
...
+export interface CategorySignDescriptor {
+    label: string
+    anchorPosition: THREE.Vector3   // World position to anchor sign to (e.g. shelf center)
+    mount: SignMount
+    style?: SignStyle
+}
+
+// ─── System ───────────────────────────────────────────────────────────────────
+
+export class CategorySignSystem {
```
EricWehrly, line 75: this isn't really for "category signs", it's maybe a generic SceneSignManager?

client/src/scene/ShelfSectionPlanner.ts
```diff
@@ -0,0 +1,136 @@
...
+        }
+        // Self-register for batch events so callers don't need to invoke onBatchPlaced
+        EventManager.getInstance().registerEventHandler(
+            StorePropsEventTypes.BatchReadyForPlacement,
+            this.handleBatchReady.bind(this)
+        )
+    }
+
+    /** Accumulate games from one batch. Called from event handler; private by design. */
+    private onBatchPlaced(games: readonly SteamGameData[]): void {
```
EricWehrly, line 50: oh no the body of `onBatchPlaced` could absolutely just be the body of `handleBatchReady` instead

client/src/scene/categorization/CategoryAssigner.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (genreA === genreB) {
+        return (b.playtime_forever ?? 0) - (a.playtime_forever ?? 0)
+    }
+
+    const idxA = KNOWN_GENRES.indexOf(genreA)
+    const idxB = KNOWN_GENRES.indexOf(genreB)
+    return idxA - idxB
+}
+
+export class CategoryAssigner {
```
EricWehrly, line 69: `CategoryAssigner` I think is fundamentally a data transformer. Let's categorize it as such (a comment is maybe fine) so that we can normalize a data transformation pattern in the codebase.

client/src/scene/categorization/CategoryAssigner.ts
```diff
@@ -0,0 +1,101 @@
...
+    private static readonly logger = Logger.createLogFunctions(CategoryAssigner.name)
+
+    assign(games: SteamGameData[]): ShelfGroup[] {
+        if (!games || games.length === 0) return []
+
+        const groupsMap = new Map<string, SteamGameData[]>()
+        const otherGroupName = 'Other'
```
EricWehrly: 'Other' absolutely needs to be an enum here.

client/src/steam/SteamApiClient.ts
```diff
@@ -294,12 +294,14 @@ export class SteamApiClient {
         steamUser: SteamUser,
         options: {
             maxGames?: number
+            /** Optional comparator to override default playtime-descending sort. */
+            sortFn?: (a: SteamGame, b: SteamGame) => number
```
EricWehrly, line 298: Instrumenting this to be able to customize the sort is perfect.
But my gut is telling me that sorting shouldn't belong here.
I thought for a bit, and this, too, I want to put in the roadmap to review.
Let's make an attempt to separate the responsibility of sorting from the steam api client ... though honestly it is a pass-through and may not be a real problem, I want to try.

client/src/core/SteamBrickAndMortarApp.ts
EricWehrly: Done — moved to `SystemUICoordinator.initializeCategoryReferencePanel()`. Net zero lines in App.

client/src/scene/categorization/CategoryAssigner.ts
EricWehrly: Agreed. `sortByGenreThenPlaytime` has a TD comment flagging it for the generic `sortByFields` utility. Noted in tech-debt.md. Sorting off `SteamApiClient` also captured there.

client/src/scene/categorization/CategoryAssigner.ts
EricWehrly: Fixed — genre assignment now uses `genre = canonical ?? otherGroupName` with a single `if (!canonical)` branch. Else blocks removed.

client/src/scene/categorization/CategoryAssigner.ts
EricWehrly: No double sort here — `shelfGroups.sort()` orders the groups themselves (Adventure shelf before RPG shelf), while `sortByGenreThenPlaytime` orders the individual games within the pipeline. They operate on different things. Added a clarifying comment.

client/src/scene/ShelfSectionPlanner.ts
EricWehrly: Done — `onBatchPlaced` is private and `ShelfSectionPlanner` self-registers for `BatchReadyForPlacement` in its constructor. And per #18, collapsed the body directly into `handleBatchReady` — no indirection.

client/src/scene/CategorySignSystem.ts
EricWehrly: Noted — leaving as tech debt (not actively updating there now). Captured in tech-debt.md.

client/src/scene/CategorySignSystem.ts
EricWehrly: Worth discussing. If it holds no instance state it could go static/module-level. Noted in tech-debt.md for the sign rendering pass.

client/src/scene/CategorySignSystem.ts
EricWehrly: Agreed on the SceneManager direction. Noted in tech-debt.md — proposed `SceneManager.get(SceneIndex)` pulling from DataManager as the canonical pattern.

client/src/scene/GpuStorePropsRenderer.ts
EricWehrly: Agreed — events should emit `Readonly<T>` on everything. Added to tech-debt.md.

client/src/scene/GpuStorePropsRenderer.ts
EricWehrly: Fixed — restored ❌ ⚠️ ✅ → ° throughout the file.

client/src/scene/GpuStorePropsRenderer.ts
EricWehrly: Fixed — degrees symbol restored in comment.

client/src/scene/GpuStorePropsRenderer.ts
EricWehrly: Fixed — checked the whole file, all emoji and special chars restored.

client/src/scene/GpuStorePropsRenderer.ts
EricWehrly: That line uses `?.` because `this.instancedShelfRenderer` is declared optional (`?`). The lines you're comparing to have the renderer already guarded before that point. Agree it could be cleaner — will tidy the guard structure in a later pass.

client/src/scene/CategorySignSystem.ts
EricWehrly: Fair point. Noted in tech-debt.md as `SceneSignManager` rename candidate — will do it when the second sign type (ceiling/recently-played) is added so the rename actually earns its keep.

client/src/scene/ShelfSectionPlanner.ts
EricWehrly: Done — collapsed into `handleBatchReady` directly. `onBatchPlaced` is gone.

client/src/scene/categorization/CategoryAssigner.ts
EricWehrly: Added a class-level doc comment: "Data transformer: maps SteamGameData[] into ShelfGroup[] bucketed by genre."

client/src/scene/categorization/CategoryAssigner.ts
EricWehrly: Done — `CategoryGroupName.Other` const-enum added and used in place of the magic string.

client/src/steam/SteamApiClient.ts
EricWehrly: Captured in tech-debt.md. Sort policy off `SteamApiClient` is the right call — the `sortFn` pass-through works for now but the responsibility is wrong. Will address when we design the groups→shelves layout conversation you described.

client/src/scene/GpuStorePropsRenderer.ts
EricWehrly: This is the later pass.

PR #39, opened 2026-04-07
Openclaw/feat recently played ceiling signs

EricWehrly:
- Track per-batch primary genre during BatchReadyForPlacement
- Compute genre cluster transitions across batch indices
- Apply alternating X offsets per cluster (left/right) at shelf creation
- Recalculate shelf bounds after all shelves are placed with final offsets
- Add focused CategoryAisleOffset unit tests for cluster + offset logic
- Cleanup CategoryAssigner CAT-DEBUG log text encoding artifacts


client/src/scene/game-box/instancing/LodGameArtworkRenderer.ts
```diff
@@ -340,7 +340,8 @@ export class LodGameArtworkRenderer {
         gameName: string,
         artworkUrl?: string,
         lodLevel: LodLevel = this.config.defaultLod,
-        highTextureSlot: number = -1
+        highTextureSlot: number = -1,
```
EricWehrly: this whole param set should really be a declared type instead

client/src/scene/game-box/instancing/LodArtworkOrchestrator.ts
```diff
@@ -274,7 +275,9 @@ export class LodArtworkOrchestrator {
                 textureIndex,
                 gameName,
                 resolvedUrl,
-                this.lazyHighTextures ? LOD_LEVEL.MID : LOD_LEVEL.HIGH
+                this.lazyHighTextures ? LOD_LEVEL.MID : LOD_LEVEL.HIGH,
+                undefined, // highTextureSlot
```
EricWehrly: let's see if https://github.com/EricWehrly/steam-brick-and-mortar/pull/39/changes#r3043565348 can get us around passing `undefined` at all here

client/src/scene/RecentlyPlayedCeilingSign.ts
```diff
@@ -0,0 +1,62 @@
...
+// Z midpoint between first arc row (r=5, z=-5) and second (r=7.8, z=-7.8) at centre angle
+const SIGN_ANCHOR_Z = -6.4
+
+/**
+ * Drop from ceiling � how far below ceiling surface the sign centre sits.
+ * Enough to be clearly readable when standing at the entrance.
+ */
+const SIGN_DROP_FROM_CEILING = 0.5
+
+export class RecentlyPlayedCeilingSign {
```
EricWehrly, line 28: This should just be "ceiling sign""

client/src/scene/RecentlyPlayedCeilingSign.ts
```diff
@@ -0,0 +1,62 @@
...
+    private readonly signSystem: SceneSignManager
+
+    constructor(scene: THREE.Scene) {
+        this.signSystem = new SceneSignManager(scene)
```
EricWehrly: Same as the comment in the other sign :/ (makes me wonder how DRY these are)

SceneSignManager can resolve scene itself from DataManager, no wiring needed.

Please extract this now, don't wait for permission. And make the similar change for scene resolution in the other code we've changed recently. Don't thread it through, look it up. Please.

client/src/scene/game-box/instancing/TextureWorker.ts
```diff
@@ -57,19 +58,10 @@ export class TextureWorker {
             this.handleWorkerMessage(event.data)
         }
         
-        this.worker.onerror = (error) => {
-            TextureWorker.logger.error('🔥 TextureWorker error:', {
```
EricWehrly, line 61: Were we not seeing this error? It feels like the right way to handle this. Did this error log not show up in tests? Was suppressed in the console?
Did it write to its own thread? I thought onerror surfaced stuff to the main thread?

client/src/scene/spawning/GameBoxSpawner.ts
```diff
@@ -98,29 +98,41 @@ export class GameBoxSpawner {
...
         shelfRotationY: number = 0
     ): void {
         const shelfSurfaces = ShelfSurfaceUtils.findShelfSurfaces(null, true)
         if (shelfSurfaces.length === 0) return
 
+        // ShelfSide naming note: 'Front' = local -Z face (away from player after arc rotation).
+        // 'Back' = local +Z face (toward origin = toward player spawn).
+        // We always populate the player-facing side (Back) first.
+        // The far side (Front) is suppressed on the back wall ring to avoid clipping.
+        // TD: rename ShelfSide.Front/Back to ShelfSide.Far/Near once naming is audited.
```
EricWehrly, line 112: "front" and "back" are sensible world-analogs to how these shelves would orient. I'm not sure why we're rename them far/near. I'm not following the logic.

I do kinda get what you're saying, but we could just have a note making it easier to understand front = near back = far?
These are 3d objects we deliberately move around and observe all sides of

client/src/scene/spawning/GameBoxSpawner.ts
```diff
@@ -98,29 +98,41 @@ export class GameBoxSpawner {
...
-                // TD: wall-shelf-back-side � wall-mounted shelves should not fill the back side.
-                // Currently all shelves fill both sides. When wall vs. floor-standing shelf types
-                // are differentiated, gate this on shelf.isWallMounted or similar.
-                const backGames = games.slice(gameIndex, gameIndex + GameLayoutConstants.GAMES_PER_SURFACE)
-                if (backGames.length > 0) {
-                    this.createGameBoxes(shelfPosition, surface, backGames, ShelfSide.Back, shelfRotationY)
-                    gameIndex += backGames.length
+                // Far side = ShelfSide.Front (-localZ, faces away from player)
+                // Suppress on back wall ring (row 4) to avoid clipping and overdensity.
+                // TD: formalize as layout policy from planner (not hardcoded row index).
```
EricWehrly, line 125: Not specifically for this, but we should highly encourage any `// TD` comments to include some kind of id or handle, so we can look up all applicable tech debt locations for a particular td item easily. 
So these would be "TD [layout]" or whatever you want to call it, because we'll want to mark all of these when we're re-evaluating layout.

client/src/scene/GpuStorePropsRenderer.ts
```diff
@@ -51,10 +51,12 @@ import { PerformanceMonitor, ASYNC_CONTEXT } from '../utils/PerformanceMonitor'
...
 import { ShelfSectionPlanner } from './ShelfSectionPlanner'
+import { RecentlyPlayedCeilingSign } from './RecentlyPlayedCeilingSign'
 import {
     computeAlternatingClusterXOffset,
     getPrimaryGenreFromBatch,
 } from './categorization/CategoryAisleOffset'
+import { computeArcShelfLayout, type ArcLayoutConfig } from './props/shared/ArcLayoutUtils'
 import type { SteamGameData } from './game-box/types/GameData'
 
 export class GpuStorePropsRenderer implements IStorePropsRenderer {
```
EricWehrly, line 62: this file continues to deffo be too long, it should be a priority to cut out the layout-related functionality. I think that should be something we try to do ahead of actually doing the layout related work. Probably a key item for our phase1-to-2-glue work...

client/src/scene/game-box/instancing/TextureWorker.ts
EricWehrly: Good catch. The worker's onerror event does fire on the main thread, but in Firefox specifically, runtime crashes inside the worker body produce an ErrorEvent where message/filename/lineno are all undefined -- so the old log of e.message showed up as literally 'undefined'. This was being swallowed visually, not actually suppressed.

Fixed in the WorkerErrorUtils refactor: extractWorkerErrorMessage checks e.error?.message first, then falls back through e.message, filename:lineno, and finally a descriptive catch-all. Tests in WorkerErrorUtils.test.ts cover the undefined-message case explicitly.

client/src/scene/GpuStorePropsRenderer.ts
```diff
@@ -258,8 +268,51 @@ export class GpuStorePropsRenderer implements IStorePropsRenderer {
...
+        // Inverted arc layout: front ring sparse (close, narrow), back rings dense (far, wide)
+        // Most-recently-played games are at the front within arm's reach;
+        // older games fill the panoramic back arcs.
+        // Per-row counts distributed to fill total: front gets fewer, back gets more.
+        // Arc config tuned so minShelfGap enforcement never reduces row counts below targets.
+        // Radii are stepped out (5.5 -> 9.5 -> 13.5 ...) and halfAngle is PI/3 for rows 0-3
+        // so each ring is wide enough to accommodate requested shelf counts at >=1m gap.
+        // Row 4 (back wall) absorbs all remaining shelves; gap enforcement is skipped for it.
+        const FIXED_ROWS_COUNT = 4 + 6 + 10 + 12  // 32 � row 0-3 fixed counts
+        const arcConfig: ArcLayoutConfig = {
```
EricWehrly, line 281: Good, but super-duper hardcoded. The hardcodedness is absolutely tech debt we need to revisit, but I'm thinking to wait until we're implementing the variable layouts work

client/src/scene/game-box/instancing/TextureWorker.ts
EricWehrly: This is great! Since we're trying to standardize the way the error handling works, would we want to instead encapsulate workers with a ManagedWorker class that instruments this onerror in a standard (but hookable) way?

client/src/scene/props/shared/ShelfSurfaceUtils.ts
```diff
@@ -31,7 +31,8 @@ export class ShelfSurfaceUtils {
             })
         })
         
-        return surfaces
+        // Sort top-to-bottom: partially-filled shelves fill from eye level down
+        return surfaces.sort((a, b) => b.topY - a.topY)
```
EricWehrly, line 35: we have tests for this?

client/src/scene/GpuStorePropsRenderer.ts
```diff
@@ -307,9 +360,9 @@ export class GpuStorePropsRenderer implements IStorePropsRenderer {
         this.shelfLayout.rows = Math.ceil(totalShelves / this.maxShelvesPerRow)
         this.shelfLayout.shelvesPerRow = this.maxShelvesPerRow
 
-        // Emit layout as soon as we know it (no need to wait for batches to finish loading)
-        // Tech debt link: docs/roadmaps/tech-debt.md → "Category System Tech Debt / Readonly event payloads"
-        EventManager.getInstance().emit<ShelfLayoutDeterminedEvent>(
+        // Emit layout once (on first allocation). Suppress during expand to avoid repeated room resizes.
+        // Tech debt link: docs/roadmaps/tech-debt.md -> "Category System Tech Debt / Readonly event payloads"
+        if (!suppressEmit) EventManager.getInstance().emit<ShelfLayoutDeterminedEvent>(
```
EricWehrly, line 365: I really don't like this change. Let's prioritize addressing it in upcoming work

client/src/utils/WorkerErrorUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Create a standardised onerror handler for a Worker.
+ * Logs the full event detail and rejects all pending requests with a meaningful error.
+ *
+ * @param workerName   - Identifier for log output (e.g. 'TextureWorker')
+ * @param pending      - Map of pending promise callbacks to reject on crash
+ * @param logger       - Optional logger object with an .error() method
+ */
+export function makeWorkerErrorHandler(
```
EricWehrly, line 40: this whole file should disappear into managedworker later

client/test/unit/scene/instancing/LodGameArtworkRenderer.test.ts
```diff
@@ -248,25 +236,25 @@ describe('LodGameArtworkRenderer', () => {
             })
             smallRenderer.initialize(mockTextureArrays, mockScene)
             
-            expect(smallRenderer.addInstance(new THREE.Vector3(), 0, 'Game 1')).toBe(0)
-            expect(smallRenderer.addInstance(new THREE.Vector3(), 1, 'Game 2')).toBe(1)
-            expect(smallRenderer.addInstance(new THREE.Vector3(), 2, 'Game 3')).toBe(-1)
+            expect(smallRenderer.addInstance({ position: new THREE.Vector3(), textureIndex: 0, gameName: 'Game 1' })).toBe(0)
```
EricWehrly, line 239: should use a type here probly?

docs/roadmaps/glue-work-before-phase2.md
```diff
@@ -0,0 +1,73 @@
...
+## Glue Work Buckets
+
+### A. Quality / Tooling (Central Focus)
+- Lint pass strategy and baseline cleanup
+- Fix high-noise formatting/encoding regressions (emoji/symbol corruption in source comments)
+- eslint rule: enforce max param count requiring type declaration (investigate `max-params` rule + config)
+- Add guardrails where we repeatedly trip (event payload mutability, panel token usage)
+
+### B. Layout and UX polish (branch merge-blocking)
+- [done] Arc layout tuning (walkability constraints, row gap enforcement)
```
EricWehrly, line 23: if its done, we aint need to track it

docs/roadmaps/glue-work-before-phase2.md
```diff
@@ -0,0 +1,73 @@
...
+- Neon "&" sign spike (see open-subagent-threads.md P3)
+- Demo branch sync (catch up openclaw/feat-demo-store to current work branch)
+- Parallel subagent workspace setup (user is copying directory for parallel branch work)
+- FOV/perspective distortion investigation (subagent investigation in progress - Apr 7)
+- Mid texture missing on some shelves (noted but not investigated - likely LOD distance issue)
+
+---
+
+## Candidate Sequence (lightweight)
+1. [in-progress] Merge-ready polish (current branch - nearly done)
```
EricWehrly, line 54: ??? No idea what "Merge-ready polish" means

PR #40, opened 2026-04-07
Openclaw/feat demo store

EricWehrly:


client/src/styles/main.css
```diff
@@ -1,4 +1,5 @@
 /* Main application styles */
+@import './ui-design-tokens.css';
```
EricWehrly: ultimately I don't like the name of `ui-design-tokens.css`. It seems too generic. We need named files to load these into, like "general-layout.css", "general-colors.css", whatever.
I also anticipate named css files for buttons, panels, etc. that create logical groupings where possible

client/src/scene/categorization/CategoryAssigner.ts
```diff
@@ -17,6 +17,36 @@ export interface ShelfGroup {
     games: SteamGameData[]
 }
 
+export type RecentlyPlayedBucket = 'today' | 'this-week' | 'this-month' | 'this-year' | 'before' | 'unplayed'
```
EricWehrly: enums innit?

client/src/scene/game-box/instancing/PixelDataCache.ts
```diff
@@ -1,25 +1,19 @@
...
- * 
- * Before (main thread IndexedDB): Cache hits took ~20-40ms, blocking rendering
- * After (worker-based): Cache hits are async with no main thread blocking
- * 
- * Architecture:
- * - Main thread: Coordinates requests, tracks stats
- * - Worker thread: All IndexedDB read/write operations
- * - ArrayBuffer transfer: Zero-copy data transfer from worker to main
- * 
- * Storage Trade-off:
```
EricWehrly, line 16: the rest made sense to remove, but this storage tradeoff was a helpful note

client/src/scene/game-box/instancing/PixelDataCache.ts
```diff
@@ -1,25 +1,19 @@
...
- * - JPEG blob: ~30-50KB per image
- * - RGBA pixels (300×450): ~540KB per image (10-18x larger)
- * - For 800 games: ~432MB IndexedDB storage
+ * PixelDataCache
+ *
+ * Web Worker-based IndexedDB cache for decoded texture pixel data.
+ * Extends ManagedWorker for standardised lifecycle and error handling.
+ *
+ * Two-phase lifecycle:
```
EricWehrly, line 7: this two-phase lifecycle is common but not universal in the codebase. We probably should consider some kind of naming or tagging convention to signal this is following that pattern.

We may also want to keep a "patterns.md".... that's some low priority tech debt tho

client/src/scene/game-box/instancing/PixelDataCache.ts
```diff
@@ -58,304 +48,191 @@ export interface CachedPixelResult {
...
     public static getInstance(): PixelDataCache {
         if (!PixelDataCache.instance) {
             PixelDataCache.instance = new PixelDataCache()
         }
         return PixelDataCache.instance
     }
-    
+
+    protected override onWorkerCrash(err: Error): void {
+        PixelDataCache.logger.error('Worker crashed:', err.message)
```
EricWehrly: these aren't generic on crash behaviors?

client/src/scene/game-box/instancing/PixelDataCache.ts
```diff
@@ -58,304 +48,191 @@ export interface CachedPixelResult {
...
-     * Initialize the worker and IndexedDB connection
+     * Initialize the worker's IndexedDB connection.
+     * Safe to call multiple times; only initializes once.
      */
     public async init(): Promise<void> {
         if (this.workerReady) return
         if (this.initPromise) return this.initPromise
-        
-        this.initPromise = this.initWorker()
+        this.initPromise = this.doInit()
```
EricWehrly: why have `init` and `doInit`?
init should init worker, which is this. return promise. No?

client/src/scene/game-box/instancing/PixelDataCache.ts
```diff
@@ -58,304 +48,191 @@ export interface CachedPixelResult {
...
-            // For PUT messages, transfer the ArrayBuffer
-            if (message.type === 'PUT') {
-                const putMessage = message as WorkerInMessage & { pixels: Uint8ClampedArray }
-                const buffer = putMessage.pixels.buffer
-                this.worker!.postMessage(message, [buffer])
-            } else {
-                this.worker!.postMessage(message)
-            }
+
+    private sendPut(message: WorkerInMessage & { pixels: Uint8ClampedArray }): Promise<PutResult> {
```
EricWehrly: added unused method?

client/src/scene/game-box/instancing/PixelDataCache.ts
```diff
@@ -58,304 +48,191 @@ export interface CachedPixelResult {
...
-        }
-        
+    ): Promise<boolean> {
+        await this.ensureReady()
         try {
-            // Make a copy since we're transferring ownership
-            const pixelsCopy = new Uint8ClampedArray(pixels)
-            
-            await this.sendMessage<PutResult>({
+            const result = await this.send<PutResult>({
```
EricWehrly: returns promise but runs await on the work being done
send returns promise, so why await result and not just return?
ok, we count errors (should be handled via error flow) and stores, and we don't seem to use stores?

client/src/scene/game-box/instancing/PixelDataCache.ts
EricWehrly, line 1: Whole file should get review from agent posing as seasoned dev

client/src/scene/game-box/GpuGameBoxRenderer.ts
```diff
@@ -205,7 +207,9 @@ export class GpuGameBoxRenderer implements IGameBoxRenderer {
         }
         
         // Priority 3: Construct portrait URL as last resort
-        if (game.appid) {
+        // Only fire if game has an artwork object — if artwork is explicitly absent,
+        // the caller wants label-only rendering (e.g. demo/test fixtures).
+        if (game.appid && game.artwork !== undefined) {
```
EricWehrly: slightly confusing prod code modification for test data. Just have the tests do an empty artwork block? They can't adapt to present games-sans-artwork under valid production code contracts?

client/src/scene/props/shared/ShelfSurfaceUtils.ts
```diff
@@ -32,6 +32,7 @@ export class ShelfSurfaceUtils {
         })
         
         // Sort top-to-bottom: partially-filled shelves fill from eye level down
+        // TD [shelf-surface-sort]: no unit test for this sort order yet - add when ShelfSurface tests exist
```
EricWehrly: just do this now tho
But mark a task for later -- utility classes should probably have a decent level of rigor in test coverage

client/src/scene/GpuStorePropsRenderer.ts
```diff
@@ -225,8 +235,64 @@ export class GpuStorePropsRenderer implements IStorePropsRenderer {
     private handleAllBatchesComplete(): void {
         this.finalizeProgressiveLoading()
         this.calculateShelfBoundsAndLayout(this.shelfPositions.length)
-        // Genre section signs skipped on recently-played branch (single group, no genre splits)\n        // Uncomment to re-enable: this.shelfSectionPlanner.planSections(this.shelfPositions)
-        this.recentlyPlayedSign.place()
+
+        const steamMode = DataManager.getInstance().get<string>(DataKey.SteamMode)
```
EricWehrly: oh, I don't like this at all. I'm not looking into the implementation just yet but like
This feels like overloading DataManager as a catch-all bucket for whatever, which ... maybe it was before, but that's still not _good_. Leaning more into it is just gonna hurt.

steamMode was obvious in the earlier changes implementing anon mode. At the end of the store load, we just went to see if we had a steam profile to load, and went into a fallback if not.

I think this is actually showing something important to align on:
cached user exists = we should load user
when user loaded, we should sort games (right now "Recently Played" but later remember user's last preference / setting ... todo for later)
Signage should respond to "on games sort".
We should also consider how layout factors in

client/src/steam-integration/SteamIntegration.ts
```diff
@@ -375,6 +385,44 @@ export class SteamIntegration {
             userInput: user.vanityUrl
         })
     }
+
+    /**
+     * Load hardcoded demo games for dev/test environments.
+     * Emits games directly into the batch pipeline — no network calls.
+     */
+    private async loadDemoGames(): Promise<void> {
```
EricWehrly, line 391: I haven't looked closely, but quick instinct is implying `loadDemoGames` belongs in `steam/fixtures/demo-games`. However much of it we need and can't manage using existing code paths....
I could be wrong, though. I'll accept justification instead of modification

client/src/scene/GpuStorePropsRenderer.ts
```diff
@@ -225,8 +236,70 @@ export class GpuStorePropsRenderer implements IStorePropsRenderer {
...
+
+        // Only show recently-played signage when we actually have recency data.
+        // Anonymous/curated stores (no rtime_last_played) skip this path entirely.
+        if (hasRecentlyPlayedData) {
+            this.recentlyPlayedSign.place()
+            this.placeTimeBucketSigns()
+        }
+    }
+
+    private placeTimeBucketSigns(): void {
```
EricWehrly: I know we've already fucked this guy with a bunch of layout-related logic ...
Why you tryna jam sign logic in? There _has_ to be one of the sign classes this fits better in

client/src/scene/NeonAmpersandSign.ts
```diff
@@ -0,0 +1,66 @@
+import * as THREE from 'three';
+
+/**
+ * NeonAmpersandSign (SPIKE / PROTOTYPE)
+ * 
+ * This is an isolated spike for a neon sign implementation using CatmullRomCurve3 and TubeGeometry.
+ * It is NOT yet wired into the main scene flow.
+ */
+export class NeonAmpersandSign {
```
EricWehrly: You said you put this on a different branch.

Guess you gaslit both of us 🤷‍♂️

client/src/scene/RecentlyPlayedCeilingSign.ts
```diff
@@ -30,13 +30,21 @@ export class RecentlyPlayedCeilingSign {
...
 
+    private position: THREE.Vector3 = new THREE.Vector3(0, RoomConstants.STORE_CEILING_HEIGHT - SIGN_DROP_FROM_CEILING, SIGN_ANCHOR_Z)
+
     constructor() {
         // SceneSignManager is a singleton - no local instance needed
     }
 
+    /** Get the position of the sign. */
+    public getPosition(): THREE.Vector3 {
+        return this.position.clone()
```
EricWehrly: clone is bad form
readonly?
clone at call site?

EricWehrly (commented):
r3047835567 - Done. Added \ShelfSurfaceUtils.test.ts\ with a top-to-bottom sort order assertion. Removed the TD comment from the source since the test now exists.

client/src/steam/fixtures/demo-games.ts
```diff
@@ -0,0 +1,56 @@
...
+function header(appid: number): string {
+    return `https://cdn.akamai.steamstatic.com/steam/apps/${appid}/header.jpg`
+}
+
+const F2P = [{ id: '37', description: 'Free to Play' }]
+const F2P_ACTION = [...F2P, { id: '1', description: 'Action' }]
+const F2P_RPG = [...F2P, { id: '3', description: 'RPG' }]
+const F2P_STRAT = [...F2P, { id: '2', description: 'Strategy' }]
+
+export const ANONYMOUS_STORE_USER: SteamUser = {
```
EricWehrly, line 26: TODO: When we figure out how to do offline access / "bookmarklet" export,
which lets us hopefully circumnavigate the lambda
(or when we resolve that as unfeasible)
let's dovetail this in and make the format of that export match our static data here

Make sure to add the dovetail to the appropriate place in roadmap/tech debt/whatever,
and let's prioritize that research task. Should be high up somewhere in subagent threads

client/src/ui/coordinators/SystemUICoordinator.ts
```diff
@@ -70,8 +70,9 @@ export class SystemUICoordinator {
             GameEventTypes.ShelfLayoutDetermined,
             (event) => {
                 const bounds = event.detail.shelfBounds
-                // Add 2m margin beyond the furthest shelf face
-                const reach = Math.abs(bounds.minZ) + 2
+                // Half the full room reach - comfortable interaction range for flatscreen.
+                // TD [raycast-drag]: don't fire if significant mouse movement (drag = look, not click).
+                const reach = (Math.abs(bounds.minZ) + 2) / 2
```
EricWehrly: So.... half and half? (halved at the comment on line 63, and you've added halfing again here)

It's counter-intuitive / misleading to not use the configured reach

And still, the reach is actually a little long. 
I want it hardcoded, but the current distance between the rings of shelves seems like a good "max reach" distance, so you should be able to borrow those numbers to arrive at the hardcode
that should be very visible and obvious when modifying the file
Act as if you'll need to come back and tune this several times, eventually for different system conditions, because that's exactly what we're going to have to do with reach

docs/README.md
```diff
@@ -51,3 +51,5 @@ Read these when touching specific complex systems (e.g., WebXR, Instancing, Even
 ### 🔬 research/ & 📦 archive/
 **Historical context.**
 Do not read these unless specifically looking for why a past decision was made or how an old feature worked.
+
```
EricWehrly: delete this whitespace so we don't have the unnecessary edited file in the PR

and try to not be this unsanitary in the future

client/src/steam-integration/SteamIntegration.ts
EricWehrly: Justification: the data (ANONYMOUS_STORE_USER, game list) already lives in steam/fixtures/demo-games.ts. What's in loadDemoGames() is the batch-emit loop — it uses EventManager, SteamEventTypes, storeSteamDataAndEmitEvent, and the yield-between-batches trick. Moving that to a fixture file would require it to import from SteamIntegration internals or duplicate the plumbing. Split is: fixtures own data, SteamIntegration owns emission. If you'd rather the batch-emit live elsewhere, the right shape would be extracting it into a shared utility that both paths call — but that's a bigger refactor. Happy to do it if you disagree.

EricWehrly (commented):
r3048445863 — Done. placeTimeBucketSigns() moved to SceneSignManager (ce601dd). GpuStorePropsRenderer now just calls SceneSignManager.instance.placeTimeBucketSigns(shelfPositions, shelfRotationsY, games, ceilingSignPos) — no sign logic remains in the renderer.

client/src/scene/game-box/instancing/TextureWorker.ts
```diff
@@ -41,56 +36,53 @@ export interface FetchAndProcessOptions {
...
-        
-        this.worker.onerror = makeWorkerErrorHandler('TextureWorker', this.pendingMessages as never, TextureWorker.logger)
+        super(TextureProcessingWorker as unknown as new () => Worker, 'TextureWorker')
+        TextureWorker.logger.lifecycle('Initialized')
+    }
 
-        this.worker.onmessageerror = (error) => {
-            TextureWorker.logger.error('TextureWorker message deserialization error:', error)
-        }
+    protected override handleMessage(data: TWOut): void {
```
EricWehrly: I don't see us using `gameNames`, just setting and deleting it. All the way down to the base class, `handleMessage` doesn't seem needed at all.

client/src/scene/game-box/instancing/texture-processing.worker.ts
```diff
@@ -251,16 +251,21 @@ ctx.onmessage = async (event: MessageEvent<WorkerMessage>): Promise<void> => {
...
+                    processed = await processBlobWithDimensions(blob, textureWidth, textureHeight, useNativeSize)
+                } else if (textureSize) {
+                    // Legacy square mode
+                    const imageData = await processBlob(blob, textureSize)
+                    processed = { imageData, width: textureSize, height: textureSize }
+                } else {
+                    // Default to native size if nothing specified
+                    processed = await processBlobWithDimensions(blob, undefined, undefined, true)
+                }
+            } catch (processError) {
```
EricWehrly: remove this change or explain what it's doing

"try
catch (error)
throw error"

isn't doing a single thing

client/src/scene/game-box/instancing/LodArtworkOrchestratorDebug.ts
```diff
@@ -13,6 +13,9 @@ import { EventManager } from '../../../core/EventManager'
 import { GameEventTypes } from '../../../types/InteractionEvents'
 import { GameArtworkProvider } from './GameArtworkProvider'
 import { DataManager } from '../../../core/data/DataManager'
+import { GpuMemoryEstimator } from '../../../debug/GpuMemoryEstimator'
```
EricWehrly, line 16: is `GpuMemoryEstimator ` one of the two memory approaches we decided to have going forward? If not, or sufficiently redundant to an existing solution, let's slag the whole thing

client/src/scene/game-box/GpuGameBoxRenderer.ts
```diff
@@ -233,6 +243,12 @@ export class GpuGameBoxRenderer implements IGameBoxRenderer {
         return { ...this.dimensions }
     }
 
+    private onAllBatchesComplete(): void {
+        this.lodDistanceManager.syncInstances()
```
EricWehrly: nope this event registration obviously goes in `lodDistanceManager` and you need to make a rule to prevent repeating this stupid mistake

PR #45, opened 2026-04-09
Openclaw/feat codebase cleanup v2

EricWehrly:


client/src/scene/props/index.ts
```diff
@@ -1,31 +1,12 @@
 /**
  * Store Props Module - Self-Contained Props System
- * 
- * This module provides a complete, self-contained system for managing store props
```
EricWehrly, line 4: why remove notes? not useful?

PR #47, opened 2026-04-10
Openclaw/feat event driven sort

EricWehrly:


docs/plans/gamesort-event-driven-plan.md
EricWehrly, line 1: we need to use the doc in clients

please dont needlessly duplicate

client/src/types/InteractionEvents.ts
```diff
@@ -198,6 +199,15 @@ export interface AllBatchesCompleteEvent extends BaseInteractionEvent {
     // Pure terminal signal: all batches are complete.
 }
 
+export interface GamesSortEvent extends BaseInteractionEvent {
```
EricWehrly: These are ... maybe EnvironmentEvents?

Certainly not InteractionEvents

client/src/scene/SignageRenderer.ts
```diff
@@ -138,10 +138,18 @@ export class SignageRenderer {
...
+     */
+    public bakeTexture(text: string, backgroundColor: number, textColor: number): THREE.DataTexture {
+        return this.createTextTexture(text, backgroundColor, textColor)
+    }
+
     /**
      * Create text texture using canvas
      */
-    private createTextTexture(text: string, backgroundColor: number, textColor: number): THREE.Texture {
+    private createTextTexture(text: string, backgroundColor: number, textColor: number): THREE.DataTexture {
```
EricWehrly, line 157: Why do we need to switch from Texture?

client/src/scene/SceneSignManager.ts
```diff
@@ -125,40 +159,91 @@ export class SceneSignManager {
...
+
+                if (this.hasRecentlyPlayedData && this.sortedGames.length > 0) {
+                    this.placeTimeBucketSignForShelf(shelfId, position as THREE.Vector3, rotY)
+                }
+            }
+        )
+
+        EventManager.getInstance().registerEventHandler(
+            GameEventTypes.GamesSort,
+            (event: CustomEvent<GamesSortEvent>) => {
```
EricWehrly: did you just define an inline method? Why?
just make a quick method for all this

client/src/scene/SceneSignManager.ts
```diff
@@ -125,40 +159,91 @@ export class SceneSignManager {
...
+
+            mesh.position.copy(signPos)
+            if (descriptor.mount.signFacingY !== undefined) {
+                mesh.rotation.y = descriptor.mount.signFacingY
+            }
+
+            return mesh
+        }
+
+        // ── Create path ───────────────────────────────────────────────────────
```
EricWehrly, line 263: this comment seems to indicate a sub-method we could break out

is doing so undesirable cuz too many params?

client/src/scene/SceneSignManager.ts
```diff
@@ -125,40 +159,91 @@ export class SceneSignManager {
...
-
+    public setSign(descriptor: CategorySignDescriptor, kind: SignKind = 'category'): THREE.Mesh {
         const style = descriptor.style ?? SignStyles.Category
         const signPos = this.resolvePosition(descriptor.anchorPosition, descriptor.mount)
+        const text = descriptor.text ?? descriptor.label
+
+        const existing = this.signs.get(descriptor.label)
+
+        if (existing) {
+            // ── Recycle path ─────────────────────────────────────────────────
```
EricWehrly, line 240: same as line 244, come to think of it

client/src/scene/categorization/GameSorter.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    constructor() {
+        EventManager.getInstance().registerEventHandler(
+            GameEventTypes.AllBatchesComplete,
+            (_event: CustomEvent<AllBatchesCompleteEvent>) => this.handleAllBatchesComplete()
+        )
+        GameSorter.logger.debug('GameSorter initialized — subscribed to AllBatchesComplete')
+    }
+
+    private handleAllBatchesComplete(): void {
```
EricWehrly: we should have a `sortByRecentlyPlayed` 

I think just straight method rename here and leave it bound as-is.

client/src/scene/categorization/GameSorter.ts
```diff
@@ -0,0 +1,87 @@
...
+        const games = DataManager.getInstance().get<SteamGameData[]>('steam.games') ?? []
+
+        if (games.length === 0) {
+            GameSorter.logger.warn('GameSorter: AllBatchesComplete fired but no games in DataManager — skipping emit')
+            return
+        }
+
+        const hasRecentlyPlayedData = games.some(g => (g.rtime_last_played ?? 0) > 0)
```
EricWehrly: Updated comment:

I don't want to do the `> 0` check that drops games without recently played data.
This means they haven't been played ("last played: never") and fall into a final category.

client/src/scene/categorization/GameSorter.ts
```diff
@@ -0,0 +1,87 @@
...
+} from '../../types/InteractionEvents'
+import type { SteamGameData } from '../game-box/types/GameData'
+import {
+    sortByRecentlyPlayed,
+    getRecentlyPlayedBucket,
+    getBucketLabel,
+    RecentlyPlayedBucket,
+} from './CategoryAssigner'
+
+export class GameSorter {
```
EricWehrly, line 64: I think I want the actual sorting logic in a kind of utility class (this class is fine)

that is then called by the organizational classes. So it's maybe the SceneLayout classes that want to initiate the sort when allBatchesComplete?

We can move to that arch later. What's here is fine for now, but I'd like to capture the intended shape

client/src/scene/shelves/ShelfRenderer.ts
```diff
@@ -0,0 +1,125 @@
...
+export class ShelfRenderer {
+    private static readonly logger = Logger.createLogFunctions(ShelfRenderer.name)
+
+    private readonly instancedShelfRenderer: InstancedShelfRenderer
+    private _isReady = false
+    /** Callbacks waiting for RendererReady to fire. */
+    private readonly readyQueue: Array<() => void> = []
+
+    constructor(maxShelfUnits = 100) {
+        this.instancedShelfRenderer = new InstancedShelfRenderer({ maxShelfUnits })
```
EricWehrly: I really do not like this.

Where is this going?
Why do we have this wrapper class? What didn't InstancedShelfRenderer have?

client/src/scene/shelves/ShelfRenderer.ts
```diff
@@ -0,0 +1,125 @@
...
+            this._isReady = true
+
+            // Drain queued waiters before emitting the event so any synchronous
+            // handlers in the queue see isReady === true immediately.
+            while (this.readyQueue.length > 0) {
+                this.readyQueue.shift()?.()
+            }
+
+            EventManager.getInstance().emit<RendererReadyEvent>(
+                StorePropsEventTypes.RendererReady,
```
EricWehrly: `this.instancedShelfRenderer.initialize` already emits this

What are you doing?

client/src/scene/shelves/ShelfLayoutCoordinator.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    public dispose(): void {
+        this.reset()
+    }
+
+    private handleFirstBatch(detail: BatchReadyForPlacementEvent): void {
+        if (!this.layoutComputed) {
+            this.layoutComputed = true
+
+            const totalShelves = detail.totalBatches
```
EricWehrly: ummm these words (shelves, batches) are not synonyms

1 "batch" of games is however many we want to resolve based on a ton of other factors

1 "shelf" of games is ultimately going to be based on how many games we want to support per shelf

which is easier currently when all shelves have a uniform # of games / game slots
When we get to wall shelves, they'll have a different arrangement and different # of games per shelf ...
(something to keep in mind, but we dont need to build right now)

client/src/scene/shelves/ShelfLayoutCoordinator.ts
```diff
@@ -0,0 +1,154 @@
...
+                shelfLayout: {
+                    rows: (shelves[shelves.length - 1]?.row ?? 0) + 1,
+                    shelvesPerRow: 10,
+                },
+            }
+        )
+
+        this.shelvesByBatch.clear()
+        for (let i = 0; i < shelves.length; i++) {
+            this.shelvesByBatch.set(i, {
```
EricWehrly, line 123: We can emit `ShelfReadyEvent` here. 

Do we want to consider a path that would not emit 40+ events back-to-back? We should evaluate with the consumers what would be best.

client/src/scene/shelves/ShelfLayoutCoordinator.ts
```diff
@@ -0,0 +1,154 @@
...
+
+            ShelfLayoutCoordinator.logger.debug(`Computing arc layout for ${totalShelves} shelves`)
+            this.computeLayout(totalShelves)
+        }
+
+        this.emitShelfForBatch(detail.batchIndex)
+    }
+
+    private computeLayout(totalShelves: number): void {
+        const arcConfig: ArcLayoutConfig = {
```
EricWehrly, line 75: most of these seem like they should be defaults inside the ArcLayout definition. We may only want to drive these where non-default...

client/src/scene/shelves/ShelfLayoutCoordinator.ts
```diff
@@ -0,0 +1,154 @@
...
+            bounds.maxZ = Math.max(bounds.maxZ, s.position.z + hd)
+        }
+
+        EventManager.getInstance().emit<ShelfLayoutDeterminedEvent>(
+            GameEventTypes.ShelfLayoutDetermined,
+            {
+                shelfBounds: bounds,
+                shelfLayout: {
+                    rows: (shelves[shelves.length - 1]?.row ?? 0) + 1,
+                    shelvesPerRow: 10,
```
EricWehrly: this is a shamelessly "magic number" _happening_ to coincide with the specified config.

Are we even using these values? We shouldn't write to event what we're not using. That's the easiest way to maintain the code.

client/src/scene/signs/TimeBucketSignHelpers.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+export function shelfBucket(
+    shelfId: number,
+    sortedGames: ReadonlyArray<Readonly<SteamGameData>>
+): RecentlyPlayedBucket | null {
+    const firstGameIndex = shelfId * SHELF_BATCH_SIZE
+    if (firstGameIndex >= sortedGames.length) return null
+
+    const bucket = getRecentlyPlayedBucket(sortedGames[firstGameIndex] as SteamGameData)
+    if (bucket === RecentlyPlayedBucket.Unplayed) return null
```
EricWehrly: Unplayed should be a valid (final) bucket for recently played

This'll basically be the same for total playtime

client/src/scene/signs/TimeBucketSignHelpers.ts
```diff
@@ -0,0 +1,90 @@
...
+export function recentlyPlayedCeilingAnchor(): THREE.Vector3 {
+    return new THREE.Vector3(0, RoomConstants.STORE_CEILING_HEIGHT - 0.5, -6.4)
+}
+
+/**
+ * Return the display label for a bucket, or null if the bucket is Unplayed.
+ * Thin wrapper kept here so callers don't need to import CategoryAssigner directly.
+ */
+export function bucketDisplayLabel(bucket: RecentlyPlayedBucket): string | null {
+    if (bucket === RecentlyPlayedBucket.Unplayed) return null
```
EricWehrly: "Never" is the appropriate label for the Unplayed bucket

client/src/types/EnvironmentEvents.ts
```diff
@@ -0,0 +1,26 @@
...
+    // Pure terminal signal: all batches are complete.
+}
+
+export interface GamesSortEvent extends BaseInteractionEvent {
+    /** Full sorted game list after all batches have loaded. */
+    sortedGames: ReadonlyArray<Readonly<SteamGameData>>
+    /** Maps bucket key (RecentlyPlayedBucket value or genre string) to human-readable label. */
+    buckets: ReadonlyMap<number | string, string>
+    /** True if any game has rtime_last_played > 0 (i.e. recently-played data is available). */
+    hasRecentlyPlayedData: boolean
```
EricWehrly: I think we shouldn't have `hasRecentlyPlayedData`

client/src/scene/categorization/GameSortFunctions.ts
```diff
@@ -0,0 +1,199 @@
...
+ * Compose comparators left-to-right. The first comparator that returns non-zero wins.
+ * Equivalent to SQL's ORDER BY a, b, c.
+ *
+ *   games.sort(chainComparators(
+ *     sortByEnumIndex('genre', KNOWN_GENRES),
+ *     sortByNumericField('playtime_forever'),
+ *     sortAlphabetically('name'),
+ *   ))
+ */
+export function chainComparators<T>(...comparators: Array<Comparator<T>>): Comparator<T> {
```
EricWehrly, line 152: this is glorious

I don't see anything, however, testing the "a then b" sort where you'd have, say, a collection of games with Action then Adventure and playtimes where Adventure was > than some of the Action, and asserting the proper order.
Those are examples, tests should make up their own 'Foo', 'Bar' genres and not rely on client code for data that's irrelevant to sorting logic

client/temp.txt
EricWehrly, line 1: this can't have been intentionally committed

client/WORK.md
EricWehrly, line 1: this can't have been intentionally committed

client/src/types/InteractionEvents.ts
```diff
@@ -409,64 +450,5 @@ export interface GameLoadingProgressEvent extends BaseInteractionEvent {
...
- * Union type of all event names for type checking
- */
-export type InteractionEventName = keyof InteractionEventMap
-
-/**
- * Utility type to get event detail type from event name
- */
-export type InteractionEventDetail<T extends InteractionEventName> = InteractionEventMap[T]
+// Map lives in EventTypeMap.ts — imported here for backward compatibility.
+export type { InteractionEventMap, InteractionEventName, InteractionEventDetail } from './EventTypeMap'
```
EricWehrly: do not re-export
we are extracting to separate concerns deliberately

client/src/types/InteractionEvents.ts
```diff
@@ -189,14 +191,30 @@ export interface ShelfLayoutDeterminedEvent extends BaseInteractionEvent {
...
+ * (e.g. layout mode switch, scene reload). Consumers that care about relayout
+ * (lighting, instanced renderers, sign placement) listen to this alongside
+ * ShelfLayoutDetermined.
+ *
+ * Payload mirrors ShelfLayoutDetermined so consumers can handle both with the
+ * same handler signature.
+ *
+ * Phase: reserved seam — no emitters exist yet. Wire behavior in the next branch.
+ */
+export interface LayoutChangedEvent extends BaseInteractionEvent {
```
EricWehrly: feels like an EnvironmmentEvent to be bro

client/src/scene/categorization/GameSorter.ts
```diff
@@ -0,0 +1,115 @@
...
+import type { SteamGameData } from '../game-box/types/GameData'
+import { sortByNumericField } from './GameSortFunctions'
+
+// Re-export so callers don't need two imports for sort + bucket types
+export { sortByNumericField, sortAlphabetically, sortByEnumIndex, chainComparators, groupByKey, groupByGenre, KNOWN_GENRES, sortByGenreThenPlaytime, resolveGenre, primaryGenre } from './GameSortFunctions'
+export type { ShelfGroup } from './GameSortFunctions'
+
+// ─── Recently-played bucket types ─────────────────────────────────────────────
+
+export enum RecentlyPlayedBucket {
```
EricWehrly, line 26: Do we really need to define this 3 times?
I don't know if we need the word "Played" in BUCKET_LABELS. Maybe we can simplify.

Also, TD note, maybe we want to carve out "recently"-specific stuff into a support file, if it is enough volume to warrant

client/src/scene/categorization/GameSorter.ts
```diff
@@ -0,0 +1,112 @@
...
+
+        if (games.length === 0) {
+            GameSorter.logger.warn('AllBatchesComplete fired but no games in DataManager — skipping emit')
+            return
+        }
+
+        // Always include all games. rtime_last_played === 0 means "never played" —
+        // those go to RecentlyPlayedBucket.Unplayed, not dropped.
+        const sortedGames: ReadonlyArray<Readonly<SteamGameData>> =
+            [...games].sort(sortByNumericField<SteamGameData>('rtime_last_played', 'playtime_forever'))
```
EricWehrly, line 86: `playtime_forever` wouldn't be doing anything here

client/src/scene/shelves/ShelfLayoutCoordinator.ts
```diff
@@ -0,0 +1,151 @@
...
+            if (totalShelves === 0) {
+                ShelfLayoutCoordinator.logger.warn('totalBatches is 0 — no shelves to lay out')
+                return
+            }
+
+            ShelfLayoutCoordinator.logger.debug(`Computing arc layout for ${totalShelves} shelves`)
+            this.computeLayout(totalShelves)
+        }
+
+        this.emitShelfForBatch(detail.batchIndex)
```
EricWehrly, line 71: this one line means we're handling every batch, not just the first one

but that's also why I added https://github.com/EricWehrly/steam-brick-and-mortar/pull/47/changes#r3066928347 as a way to work that out

PR #50, opened 2026-04-11
Openclaw/feat linter wip

EricWehrly:


client/src/lighting/ManagedLights.ts
```diff
@@ -8,6 +8,7 @@ interface IManagedLight {
     addToParent(parent: THREE.Object3D, scene: THREE.Scene, name?: string): this
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 function managedLightMixin<T extends new (...args: any[]) => THREE.Light>(Base: T, lightType: string) {
```
EricWehrly, line 12: Why can't we just get away with 'unknown' here?

client/src/steam/cache/CacheManager.ts
EricWehrly, line 1: It looks like this entire file is unused in favor of its SimpleCacheManager sibling.

Better to just remove?

client/src/steam/cache/SimpleCacheManager.ts
```diff
@@ -36,6 +36,7 @@ export interface CacheStats {
  * Generic caching layer that can wrap any method with caching functionality
  */
 export class CacheManager {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     private cache: Map<string, CacheEntry<any>> = new Map()
```
EricWehrly: same q here as above -- 'unknown' isn't the intended specificity?

client/src/utils/textures/patterns/BasePatternGenerator.ts
```diff
@@ -45,7 +45,7 @@ export abstract class BasePatternGenerator {
       width = 512,
       height = 512,
       seed = Date.now(),
-      backgroundColor = '#8B0000'
+      _backgroundColor = '#8B0000'
```
EricWehrly: this didn't need an `_`. Deletion was the cleaner fix. Done locally.

client/src/debug/GameFinder.ts
```diff
@@ -60,11 +60,10 @@ export class GameFinder {
...
 
     private constructor() {
-        const scene = DataManager.getInstance().get<THREE.Scene>(DataKey.MainScene)
-        if (!scene) {
-            throw new Error('GameFinder requires scene to be registered in DataManager')
-        }
-        this.scene = scene
+        this.scene = DataManager.getInstance().getOrThrow<THREE.Scene>(
+            DataKey.MainScene,
+            'GameFinder requires scene to be registered in DataManager'
```
EricWehrly: If we drop the custom message, I _think_ we still get the same info
1. `scene` (DataKey.MainScene at least) is logged as the key in the internal DataManager throw
2. GameFinder should surface in the call stack. 
Arguably makes these 3 lines a one-liner, if slightly dense.

client/src/scene/RoomManager.ts
```diff
@@ -66,17 +66,15 @@ export class RoomManager {
...
-        
-        const camera = DataManager.getInstance().get<THREE.PerspectiveCamera>(DataKey.MainCamera)
-        if (!camera) {
-            throw new Error('RoomManager requires camera to be registered in DataManager')
-        }
-        this.camera = camera
+        const dataManager = DataManager.getInstance()
+        this.scene = dataManager.getOrThrow<THREE.Scene>(
+            DataKey.MainScene,
+            'RoomManager requires scene to be registered in DataManager'
```
EricWehrly: same as https://github.com/EricWehrly/steam-brick-and-mortar/pull/50/changes#r3068370738

client/src/utils/textures/patterns/GeometricPatternGenerator.ts
```diff
@@ -19,11 +19,6 @@ export class GeometricPatternGenerator extends BasePatternGenerator {
     const geometricOptions = options as GeometricPatternOptions
     const {
       variant = 'standard',
-      glowIntensity = 8,
-      shapeCount,
-      shapeTypes = ['circle', 'triangle', 'lightning', 'diamond'],
```
EricWehrly, line 24: can we drop a 'todo' to support the different shape types?
We probably won't do it, but I want to keep a pulse on it

client/src/lighting/ManagedLights.ts
EricWehrly: I tested replacing `any[]` with `unknown[]` in this mixin constraint, and TypeScript broke subclass constructor/static typing (managed spot/rect lights started losing typed members like `target` and `shadow`).

This boundary is constructor forwarding only (`new (...args) => THREE.Light`); we never inspect args. So `any[]` here is intentionally narrow to preserve constructor compatibility, not a general escape hatch.

client/src/steam/cache/CacheManager.ts
EricWehrly: Yep — agreed. I removed `client/src/steam/cache/CacheManager.ts` in `ff8eab6`. `SteamApiClient` is using `SimpleCacheManager`, and I couldn't find any remaining imports/usages of the old file.

PR #52, opened 2026-04-13
Openclaw/feat steamspy hydrator

EricWehrly:


external-tool/infrastructure/modules/lambda-hydrator/main.tf
```diff
@@ -0,0 +1,115 @@
...
+}
+
+# Lambda function
+resource "aws_lambda_function" "steamspy_hydrator" {
+  filename         = data.archive_file.lambda_zip.output_path
+  function_name    = "${var.project_name}-${var.environment}-steamspy-hydrator"
+  role            = aws_iam_role.lambda_execution.arn
+  handler         = "index.handler"
+  source_code_hash = data.archive_file.lambda_zip.output_base64sha256
+  runtime         = "nodejs20.x"
```
EricWehrly: we should use a newer node
newest LTS AWS supports

external-tool/infrastructure/lambda-hydrator-src/index.js
```diff
@@ -0,0 +1,300 @@
...
+    // If we have a nested data.full_data structure (typical appdetails response), reconstruct it safely
+    if (baseData.data && baseData.data.full_data) {
+      finalData = {
+        ...baseData,
+        data: {
+          ...baseData.data,
+          full_data: mergedFullData
+        }
+      };
+    } else {
```
EricWehrly: I don't want us to do this, mergedFullData seems correct

external-tool/infrastructure/lambda-hydrator-src/index.js
```diff
@@ -0,0 +1,300 @@
...
+      finalData = mergedFullData;
+    }
+    console.log(`[AppID ${appid}] Successfully merged SteamSpy data.`);
+  } else {
+    console.log(`[AppID ${appid}] Writing unhydrated base data as fallback.`);
+  }
+
+  // 4. Write back to S3 hydrated path
+  const targetKey = `appDetailsWithTags/${appid}.json.gz`;
```
EricWehrly: I think I want to drive `appDetailsWithTags` in from the environment, so we can customize it in the lambda's environment variables

external-tool/infrastructure/lambda-hydrator-src/index.js
```diff
@@ -0,0 +1,300 @@
...
+      if (match) {
+        keys.add(match[1]);
+      }
+    });
+    continuationToken = res.NextContinuationToken;
+  } while (continuationToken);
+  return keys;
+}
+
+exports.handler = async (event, context) => {
```
EricWehrly: let's put this at the top of the file

external-tool/infrastructure/lambda-hydrator-src/index.js
```diff
@@ -0,0 +1,300 @@
...
+        body: JSON.stringify(result)
+      };
+    } catch (error) {
+      console.error(`[AppID ${event.appid}] Hydration failed critically:`, error);
+      return {
+        statusCode: 500,
+        body: JSON.stringify({ message: error.message, stack: error.stack })
+      };
+    }
+  }
```
EricWehrly: this should be a handler submethod, as should encapsulate the "automated batch sweep"

external-tool/infrastructure/lambda-hydrator-src/index.js
```diff
@@ -0,0 +1,300 @@
...
+    console.log("Another hydrator instance is currently running. Exiting to prevent Rate Limit conflicts.");
+    return {
+      statusCode: 429,
+      body: JSON.stringify({ message: "Job is already locked and running." })
+    };
+  }
+
+  try {
+    // 2. Discover missing hydrated apps
+    console.log("Listing base keys from 'appdetails/'...");
```
EricWehrly: we know this works, so we don't need to commit these debug logs
we are paying for these logs at some scale, so let's not include debug/trace statements

external-tool/infrastructure/lambda-hydrator-src/index.js
```diff
@@ -0,0 +1,300 @@
...
+    const baseKeys = await listAllKeys('appdetails/');
```
EricWehrly: if we don't include a prefix, and deal with one after, we make one external call rather than 2

... and we do need to consider pagination either way, right? I think that's todo for next phase?

external-tool/infrastructure/lambda-hydrator-src/index.js
```diff
@@ -0,0 +1,316 @@
...
+    const res = await s3Client.send(listCmd);
+    (res.Contents || []).forEach(item => {
+      keys.add(item.Key);
+    });
+    continuationToken = res.NextContinuationToken;
+  } while (continuationToken);
+  return keys;
+}
+
+async function runAutomatedBatchSweep(context) {
```
EricWehrly, line 188: we can still carve this up into submethods for readability

the comments are basically guiding it

external-tool/infrastructure/lambda-hydrator-src/index.js
```diff
@@ -0,0 +1,316 @@
...
+      Bucket: CACHE_BUCKET_NAME,
+      Key: sourceKey,
+    });
+    const { Body } = await s3Client.send(getCommand);
+    const compressedData = await streamToBuffer(Body);
+    baseData = JSON.parse(gunzipSync(compressedData).toString('utf-8'));
+  } catch (err) {
+    console.error(`[AppID ${appid}] Failed to read base data from S3 (${sourceKey}):`, err);
+    throw err;
+  }
```
EricWehrly, line 56: func getBaseDataFromS3

external-tool/infrastructure/modules/lambda-hydrator/main.tf
EricWehrly: I think we should do nodejs22
and while we're at it, run in arm64. should be cheaper
maybe we should go flip the other lambda to arm64 and node22 as well, see if they still work alright 
level em up while we're monkeying

PR #54, opened 2026-04-13
Openclaw/feat trigger hydrator

EricWehrly:


external-tool/infrastructure/lambda-src/services/cache.js
```diff
@@ -40,27 +40,39 @@ async function getFromCache(appid) {
...
     const response = await s3Client.send(command);
     const compressed = await streamToBuffer(response.Body);
     const decompressed = gunzipSync(compressed);
-    const data = JSON.parse(decompressed.toString('utf-8'));
-    
-    // Populate memory cache
+    return JSON.parse(decompressed.toString('utf-8'));
+  };
+
+  try {
```
EricWehrly, line 52: We repeat these try blocks. Maybe it makes sense to do the heavier try lifting down in the fetch method, and just return null if the fetch goes wrong?

external-tool/infrastructure/lambda-src/package.json
```diff
@@ -11,13 +11,14 @@
     "start": "yarn local-server"
   },
   "dependencies": {
-    "@aws-sdk/client-secrets-manager": "^3.450.0",
```
EricWehrly, line 14: are we sure about removing secrets-manager?

PR #60, opened 2026-04-14
feat: integrate SteamSpy metadata and enable rating sort

EricWehrly:
Pulls hydrated SteamSpy fields through the data layer, updates the normalizer, and exposes a new 'By User Rating' sorting layout. Sort uses user score and falls back to playtime for ties.


client/src/scene/categorization/GameSorter.ts
```diff
@@ -205,6 +234,40 @@ export class GameSorter {
...
+        const buckets = new Map<string, string>()
+        for (const game of sortedGames) {
+            const score = game.userscore ?? 0
+            
+            let bucketKey: string
+            let label: string
+            
+            if (score >= 90) {
+                bucketKey = 'overwhelmingly-positive'
+                label = 'Overwhelmingly Positive (90%+)'
```
EricWehrly: lets drop the parentheses from this
would that let us more cleanly express label thresholds as a map rather than needing this whole if/else/else/else block?

client/src/scene/categorization/GameSorter.ts
```diff
@@ -166,6 +167,34 @@ export class GameSorter {
         GameSorter.logger.debug(`GamesSort emitted (by playtime): ${sortedGames.length} games`)
     }
 
+    public sortByRating(): void {
```
EricWehrly, line 169: I thought we had established a generic numeric sort that precluded us needing to create these specific sort functions?

PR #62, opened 2026-04-14
feat(steam): introduce forceRefresh flag to bypass cache safely on ex…

EricWehrly:
…plicit request


client/src/steam-integration/SteamIntegration.ts
```diff
@@ -99,14 +100,16 @@ export class SteamIntegration {
...
+            dataManager.set('steam.userId', steamId, {
                 domain: DataDomain.SteamIntegration
             })
+        } else {
+            SteamIntegration.logger.warn('No steamid found in game library state; cannot update DataManager.userId')
         }
         
         this.eventManager.emit<SteamDataLoadedEvent>(SteamEventTypes.DataLoaded, {
-            userInput
+            userInput: steamId || 'unknown'
```
EricWehrly: we can drop the parameter entirely from the event
the event itself doesn't get consumed, but we'll keep it
We can drop the param from it though

client/src/steam-integration/SteamIntegration.ts
```diff
@@ -200,17 +203,14 @@ export class SteamIntegration {
...
+            async refreshData(callbacks: ProgressCallbacks = {}, ignoreCache = false, targetUser?: string): Promise<GameLibraryState | null> {
+        if (!targetUser) {
+            callbacks.onStatusUpdate?.('No target user specified to refresh', 'error')
             return null
         }
         
-        callbacks.onStatusUpdate?.('🔄 Reloading data...', 'loading')
-        return this.loadGamesForUser(currentState.userData.vanity_url, callbacks)
+        callbacks.onStatusUpdate?.('?? Reloading data...', 'loading')
+        return this.loadGamesForUser(targetUser, callbacks, ignoreCache)
```
EricWehrly: isnt this just gonna pass null and have issues?

PR #63, opened 2026-04-15
fix(draw-calls): name all unnamed Mesh objects for drawCallReport rea…

EricWehrly:
…dability

All unnamed Mesh objects now have descriptive names so drawCallReport() output is actionable rather than a wall of "(unnamed)".

Changes:
- SignageRenderer.createSign(): sign-canvas-{text-slug} (canvas signs are the dominant source of unnamed meshes — one per shelf bucket/section sign)
- NeonTubeSignRenderer: neon-tube-{id}-{segment-index} per tube segment
- BlockLetterSignRenderer: sign-block-letter-{id} per letter mesh
- PropRenderer: wire-rack-post-N, wire-rack-wire-{left|right|back}-N, category-divider-{post|cap}, floor-marker-{center|left|right}

Tests added:
- SignageRenderer: asserts sign-canvas- prefix and unique names per text
- PropRenderer: asserts zero unnamed meshes for each prop factory method

Note on instancing: canvas signs cannot be instanced (each has a unique texture baked from its text). Neon tubes and block letters could share geometry in theory but that's a larger refactor. Tracked separately.


docs/tech-debt.md
```diff
@@ -32,7 +32,19 @@
 
 ---
 
-## id: system-events-split
```
EricWehrly, line 35: you just killed this line, instead of displacing it

PR #74, opened 2026-04-17
Openclaw/first run transition

EricWehrly:


client/test/unit/steam-integration/anonymous-store-flow.test.ts
```diff
@@ -30,10 +32,15 @@ describe('Anonymous Store Data Storage (Bug Fix)', () => {
 
     it('should store demo games in DataManager after proper initialization sequence', async () => {
         // Setup: Replicate what loadDemoGames() does
+        const makeGame = (appid: number, name: string, playtime: number): SteamGame => ({
```
EricWehrly, line 35: is a function really necessary? I wanted to move to defining the games in a json file so it's even cleaner
this makes that harder

client/src/steam/batch/BatchAppDetailsClient.ts
```diff
@@ -38,6 +38,13 @@ export interface BatchAppDetailsOptions {
      * Callback for batch progress
      */
     onProgress?: (fetched: number, total: number) => void;
+
+    /**
+     * Callback invoked after each network batch completes with that batch's results.
+     * Enables progressive emission — callers can process each batch as it arrives
+     * rather than waiting for all batches to finish.
+     */
+    onNetworkBatchReady?: (batchIndex: number, totalBatches: number, results: Map<number, AppDetailsResponse>) => Promise<void> | void;
```
EricWehrly: do we not already have a "batchready" event? Why should network be different?

PR #76, opened 2026-04-18
feat(game-boxes): split load/place in GpuGameBoxRenderer for re-sort …

EricWehrly:
…support

Phase 1 (BatchReadyForPlacement): prewarmGame() prefetches artwork into atlas
Phase 2 (GamesSort): clearPlacements() + placeGame() assigns sorted positions

- LodGameArtworkRenderer: add clearPlacements() to reset instances without freeing texture slots
- LodArtworkOrchestrator: add prefetchArtwork(), placeInstance(), clearPlacements() + prefetchedArtworkUrl map
- ILodArtworkRenderer: expose prefetchArtwork, placeInstance, clearPlacements on interface
- GpuGameBoxRenderer: add prewarmGame(), placeGame(), clearPlacements(); remove GameBoxSpawned subscription
- InstancedLabelRenderer: add clear() for placement reset without texture disposal
- GameBoxSpawner: rewrite to two-phase model; subscribe to GamesSort for placement; expose setRenderer() so GpuStorePropsRenderer can inject the renderer after deferred construction
- GpuStorePropsRenderer: call setRenderer() after renderer creation; null it on clearProps()
- GameBoxSpawner.test.ts: rewrite tests to match two-phase contract


client/docs/plans/label-renderer-simplification.md
EricWehrly, line 1: this is leftover from previous work and shouldn't be needed

is it not done now?

docs/plans/first-run-status.md
EricWehrly, line 1: this was a temporary document and is probably no longer needed

client/src/scene/game-box/instancing/LodArtworkOrchestrator.ts
```diff
@@ -231,6 +235,126 @@ export class LodArtworkOrchestrator implements ILodArtworkRenderer {
...
 
+    /**
+     * Phase 1 of the load/place split: fetch and cache artwork for a game without
+     * placing a GPU instance. Safe to call as batches arrive, before shelf positions
+     * are known. Idempotent — calling again for the same appId is a no-op.
+     *
+     * @returns 'prefetched' when texture was freshly loaded, 'cached' if already done,
+     *          'permanent-failure' if the artwork URL is known-bad, 'error' on unexpected failure.
+     */
+    public async prefetchArtwork(
```
EricWehrly, line 310: could we chop this up? or down? :)

client/src/scene/game-box/GpuGameBoxRenderer.ts
```diff
@@ -120,27 +118,68 @@ export class GpuGameBoxRenderer implements IGameBoxRenderer {
...
+     * Call as batches arrive. Idempotent — subsequent calls for the same game are no-ops.
+     * Permanent artwork failures are tracked here; the label box is created in placeGame()
+     * once a world position is available.
+     */
+    public async prewarmGame(game: SteamGameData, side: ShelfSide = ShelfSide.Front): Promise<void> {
+        const artworkUrl = this.selectBestArtworkUrl(game)
+        const appid = typeof game.appid === 'number' ? game.appid : 0
+
+        if (!artworkUrl) {
+            if (AppSettings.get(Setting.EnableLabels)) {
```
EricWehrly: This check isn't repeated in the catch below.
Maybe we should perform the check when `pendingLabelGames` would be used (or later), and let the array be no-op.
it localizes the check better if we instead initialize `pendingLabelGames` as a "/dev/null" instead, so these calls can be treated cleanly (the "Feature Flag" (user option) doesn't need to repeat around the codebase, and interfacing structures can treat it as normal...)

PR #78, opened 2026-04-19
Openclaw/force cache update

EricWehrly:


client/src/steam/SteamApiClient.ts
```diff
@@ -152,7 +168,8 @@ export class SteamApiClient {
                 SteamApiClient.logger.warn('User has 0 games — may indicate privacy settings or an empty library')
             }
 
-            this.cache.set(cacheKey, response)
+            // 48 hour TTL for game lists
+            this.cache.set(cacheKey, response, { ttlMs: 48 * 60 * 60 * 1000 })
```
EricWehrly: extract this value to where we _could_ drive it with config if we want to, but let's leave it (I think) in this file for now
please

PR #81, opened 2026-04-19
Openclaw/feat stock strategy

EricWehrly:


client/src/scene/props/shared/StockStrategy.ts
```diff
@@ -0,0 +1,80 @@
...
+ * ArcStockStrategy
+ *
+ * Default strategy for arc shelf layouts.
+ * Fills all Near faces top-to-bottom first, then all Far faces as overflow.
+ * This gives a contiguous readable face before wrapping to the back side.
+ *
+ *   board[0].near, board[1].near, board[2].near,
+ *   board[0].far,  board[1].far,  board[2].far
+ */
+export class ArcStockStrategy implements IStockStrategy {
```
EricWehrly: I think these probably make more sense in their respective utils files

client/src/scene/props/StorePropsCoordinator.ts
```diff
@@ -159,6 +178,43 @@ class StorePropsCoordinator {
...
+        this.lastTotalBatches = 0
+
+        // Build fresh subsystems under the new layout mode.
+        this.eventManager.emit<StorePropsSetupRequestEvent>(StorePropsEventTypes.SetupRequest, {
+            source: EventSource.System,
+        })
+
+        // Re-emit LoadLibrary so the batch pipeline runs under the new layout.
+        // SteamIntegration will emit ClearRequest internally, which we suppress via the flag.
+        import('../../core/data').then(({ DataManager }) => {
```
EricWehrly: I would prefer the linter forbade us inline imports like this
So
1. use a proper top-of-file import
2. see about adding a lint rule that would discourage this

client/src/scene/shelves/ShelfLayoutCoordinator.ts
```diff
@@ -84,35 +102,26 @@ export class ShelfLayoutCoordinator {
...
-                Math.PI / 3.5,
-                Math.PI / 3,
-                Math.PI / 3,
-                Math.PI / 2.6,
-            ],
-            minShelfGap: 1.0,
-            rowRadiusStep: 4.0,
-            firstRowRadius: 5.5,
+    private static strategyFor(layoutMode: LayoutMode): IStockStrategy {
+        switch (layoutMode) {
```
EricWehrly: we should keep an eye on this, as it's 1:1 with the utils right now, and doesn't make sense to know about the strategy implementations here

but it depends on how the strategy evolves...

client/src/scene/shelves/ShelfLayoutCoordinator.ts
```diff
@@ -84,35 +102,26 @@ export class ShelfLayoutCoordinator {
...
+            case 'spoke': return new SpokeStockStrategy()
+            case 'row':   return new RowStockStrategy()
+            case 'arc':
+            default:      return new ArcStockStrategy()
         }
+    }
 
-        const shelves = computeArcShelfLayout(totalShelves, arcConfig)
+    private computeLayout(totalShelves: number): void {
+        const shelves = this.layoutMode === 'spoke'
```
EricWehrly: this is a terrible way to start this case/mapping

client/src/scene/shelves/ShelfLayoutCoordinator.ts
```diff
@@ -133,16 +142,53 @@ export class ShelfLayoutCoordinator {
...
+                row: 'row' in s ? (s as ArcShelfInfo).row : i,
+                indexInRow: 'indexInRow' in s ? (s as ArcShelfInfo).indexInRow : i,
             })
         }
 
-        ShelfLayoutCoordinator.logger.debug(`Layout determined for ${shelves.length} shelves`) 
+        ShelfLayoutCoordinator.logger.debug(`Layout determined for ${shelves.length} shelves`)
+    }
+
+    private computeArcLayout(totalShelves: number): ArcShelfInfo[] {
```
EricWehrly, line 152: this absolutely should be in the arc layout file
for which "utils" may be a misleading suffix...

client/src/scene/spawning/GameBoxSpawner.ts
```diff
@@ -231,7 +233,7 @@ export class GameBoxSpawner {
                 const batchIndex = sortedBatchIndices[batchIndexCursor]
                 const shelfPos = this.shelfPositions.get(batchIndex)!
                 const stockSurfaces = GameBoxUtils.buildStockSurfaces(
-                    shelfPos.position, shelfPos.rotationY, shelfSurfaces
+                    shelfPos.position, shelfPos.rotationY, shelfSurfaces, undefined, this.stockStrategy
```
EricWehrly: (why) undefined?

With 5 parameters, maybe this input needs to be a type
if we want optional parameters to be undefined, should they be re-arranged to be last?

client/src/scene/props/StorePropsCoordinator.ts
```diff
@@ -159,6 +179,41 @@ class StorePropsCoordinator {
         StorePropsCoordinator.logger.debug('Entrance mat placed at origin')
     }
 
+    private handleLayoutRequested(detail: LayoutRequestedEvent): void {
+        if (detail.layoutMode === this.activeLayoutMode) return
+        this.activeLayoutMode = detail.layoutMode
+        StorePropsCoordinator.logger.info(`Layout mode → ${detail.layoutMode}; triggering scene rebuild`)
+
+        // Guard against the ClearRequest that SteamIntegration emits mid-reload
```
EricWehrly, line 187: this is definitely tech debt to come back to at some point

PR #84, opened 2026-04-20
Openclaw/feat section per layout

EricWehrly:


client/src/scene/spawning/GameBoxSpawner.ts
```diff
@@ -130,104 +119,113 @@ export class GameBoxSpawner {
...
-            }
-            const estimatedGames = totalBatches * 18
-            this.renderer = new GpuGameBoxRenderer(estimatedGames + 100)
-            this.rendererInitialized = true
-            this.lastTotalBatches = totalBatches
+        // Construct renderer on first batch using total game estimate.
+        // Renderer is replaced if batch count changes (library reload).
+        if (!this.renderer || totalBatches * 18 + 100 > (this.renderer as any)._capacity) {
+            this.renderer?.dispose()
+            this.renderer = new GpuGameBoxRenderer(totalBatches * 18 + 100)
```
EricWehrly, line 154: we probably want to put some tech debt around this naive assumption of games length that we probably don't need

client/src/scene/batch/BatchCoordinator.ts
```diff
@@ -222,6 +222,21 @@ export class BatchCoordinator<T> {
             this.isProcessing = false
         }
 
+        // All batch data has been dispatched — games are in DataManager and all
+        // BatchReadyForPlacement events have fired. GameSorter can now sort and
+        // plan sections. This is earlier than AllBatchesComplete, which waits
+        // for GamesPlaced confirmation.
+        if (this.received > 0 && this.received === this.expectedTotal) {
+            const totalGames = this.queue.length === 0
```
EricWehrly, line 230: I don't like us emitting a guestimate. Especially when we could just get total count from the datamanager if we needed it. 
Does the event or any subscribers need it?

PR #88, opened 2026-04-22
Openclaw/feat section per layout v2

EricWehrly:


client/src/ui/binder/BinderGameDetailPanel.ts
```diff
@@ -103,6 +111,22 @@ export class BinderGameDetailPanel {
         }
     }
 
+    private formatRating(userscore: number): string {
```
EricWehrly: okay now I know we already have this mapping somewhere in our data-like layer

client/src/steam/GamesLoader.ts
```diff
@@ -53,6 +57,12 @@ export class GamesLoader {
         const refreshBatchCount = Math.ceil(refreshAppids.length / BATCH_SIZE)
         const totalBatchCount = renderableBatchCount + refreshBatchCount
 
+        EventManager.getInstance().emit<SteamLibraryManifestReadyEvent>(SteamEventTypes.LibraryManifestReady, {
+            totalGames: sortedGames.length,
+            totalBatches: totalBatchCount,
```
EricWehrly: are we using totalBatches? Are we using all our params?

if we're sending an appids collection, we don't need to provide totalGames since it can be derived from appids.length. 
but if appids isn't actually getting used, as seems to be the case when I investigate the code, then we can drop that and just keep totalgames...

client/src/scene/ShelfSectionPlanner.ts
```diff
@@ -156,14 +199,55 @@ export class ShelfSectionPlanner {
...
     public dispose(): void {
         this.signSystem.dispose()
     }
+
+    /**
+     * Debug helper: place a sign on every known shelf position showing its
+     * shelf index and section index. Useful for diagnosing spoke layout geometry.
+     * Call from browser console: `window.__debugSectionPlanner?.labelAllShelves()`
+     */
+    public labelAllShelves(): void {
```
EricWehrly: we didn't turn out to need this,
and I don't want to spend the time verifying it works,
so we should just remove it
we can re-add if we turn out to need it

client/src/scene/SceneCoordinator.ts
```diff
@@ -78,6 +78,9 @@ export class SceneCoordinator {
         this.gameSorter = new GameSorter()
         this.sceneSignManager = SceneSignManager.instance
         this.shelfSectionPlanner = new ShelfSectionPlanner()
+        if (window) {
```
EricWehrly: don't do this, please

client/src/scene/categorization/GameSorter.ts
```diff
@@ -72,11 +76,18 @@ export class GameSorter {
...
-        // Re-apply current modes on subsequent loads (layout switch, library reload)
         this.arrange(this.activeGroupMode, this.activeSortMode)
     }
 
     private handleArrangementRequested(detail: ArrangementRequestedEvent): void {
+        // Signal to GPU renderers that geometry placements will be rebuilt.
+        // This ensures stale shelf instances from prior arrangement are cleared
+        // before new sections/shelf positions are computed.
+        // LayoutClearRequest resets placement state but leaves GPU atlas intact.
+        EventManager.getInstance().emit<StorePropsLayoutClearRequestEvent>(
```
EricWehrly: why not just have the listening classes listen to the ArrangementRequestedEvent instead?

PR #141, opened 2026-07-14
Act2/desktop local data pipeline

EricWehrly:


client/src/steam/LocalSteamDataWriter.ts
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        const metadata = await invoke<LocalAppMetadata[]>('read_local_app_metadata', { appids })
+
+        // Wait for the baked-cache seed to have already landed (or failed) before touching
+        // AppDetailsCache ourselves. Both this write and the seed start around app startup with
+        // no ordering guarantee between them; without this wait, this write can land first and
+        // "win" the race, so existingArtwork below sees nothing to preserve even though the seed
+        // was about to provide real data for the same appids a moment later.
+        await SteamApiClient.getInstance().waitForAppDetailsCacheSeed()
```
EricWehrly: This is exactly what we have events for. to avoid needing to reach into other classes.

client/src/steam/GamesLoader.ts
```diff
@@ -108,6 +108,40 @@ export class GamesLoader {
...
 
+    /**
+     * Fetches appdetails for the given appids directly from the network and writes them into
+     * AppDetailsCache - no cache check first (caller already knows these aren't cached), no
+     * BatchEmitter streaming (this is a one-off gap-fill, not the main progressive-load path).
+     * Used by LocalSteamLibraryLoader to resolve collection-referenced appids the local scan has
+     * no appinfo.vdf data for. A per-appid fetch failure is silently omitted from the returned
+     * map rather than failing the whole call - same tolerance as fetchAndEmitUncached.
+     */
+    public async fetchAndCacheAppDetails(appids: number[]): Promise<Map<number, AppDetailsData>> {
```
EricWehrly, line 125: Wait why arent we using fetchandemituncached? to avoid emit? why do we need to?

client/src/ui/TaxonomyOptionAvailability.ts
```diff
@@ -0,0 +1,57 @@
...
+function hasAnyTags(game: SteamGameData): boolean {
+    if (game.steamspy_top_tags && game.steamspy_top_tags.length > 0) return true
+    return getTopSteamSpyTags(game.steamspy_tags).length > 0
+}
+
+export function computeAvailableDimensions(games: readonly SteamGameData[]): AvailableDimensions {
+    const groupModes = new Set<GroupMode>([GroupModes.None, GroupModes.ByPlaytime])
+    const sortModes = new Set<SortMode>([SortModes.Alphabetical, SortModes.ByPlaytime])
+
+    for (const game of games) {
```
EricWehrly: We have reducers that would do this more cleanly

desktop/tauri-app/src/steam/playtime.rs
```diff
@@ -0,0 +1,141 @@
...
+        .and_then(|v| v.as_obj())
+        .ok_or_else(|| "localconfig.vdf missing Software/Valve/Steam/apps block".to_string())?;
+
+    let mut playtimes = Vec::new();
+    for (appid_str, entry) in apps {
+        let Ok(appid) = appid_str.parse::<u32>() else {
+            continue; // skip non-numeric keys if any ever show up here
+        };
+        let last_played = entry
+            .get("LastPlayed")
```
EricWehrly: is there any way for us to define a structure or something for these object definitions? Rather than passing them through straight strings?

client/src/steam/LocalSteamLibraryLoader.ts
```diff
@@ -0,0 +1,177 @@
...
+ */
+async function resolveRemainingAppidsFromNetwork(candidateAppids: ReadonlySet<number>): Promise<void> {
+    const appDetailsCache = new AppDetailsCache()
+    const missingAppids = await appDetailsCache.findMissing([...candidateAppids])
+    if (missingAppids.length === 0) {
+        return
+    }
+
+    try {
+        const resolved = await SteamApiClient.getInstance().fetchAndCacheAppDetails(missingAppids)
```
EricWehrly: If we have to use this, I'd much prefer just call gamesloader rather than needing to write a pass-through in steamapiclient

client/src/steam-integration/SteamIntegration.ts
```diff
@@ -345,7 +365,7 @@ export class SteamIntegration {
             this.storeSteamDataAndEmitEvent(library.owner.displayName ?? 'imported-library')
             await this.emitGamesInBatches(enrichedGames)
 
-            if (library.owner.steamId && AppSettings.get('autoLoadProfile')) {
+            if (library.owner.steamId && library.provenance.channel !== 'local-scan' && AppSettings.get('autoLoadProfile')) {
                 SteamIntegration.logger.info(`Background re-fetch for steamId ${library.owner.steamId}`)
                 this.eventManager.emit<SteamLoadLibraryEvent>(SteamEventTypes.LoadLibrary, {
```
EricWehrly: this event emission is actually confusing and extra boilerplate since it's just consumed by a method in this class that should be called directly.

Events are the preferred way to traverse classes, WAY overkill for this "call a method in this class"

client/src/steam-integration/SteamIntegration.ts
```diff
@@ -345,7 +365,7 @@ export class SteamIntegration {
             this.storeSteamDataAndEmitEvent(library.owner.displayName ?? 'imported-library')
             await this.emitGamesInBatches(enrichedGames)
 
-            if (library.owner.steamId && AppSettings.get('autoLoadProfile')) {
+            if (library.owner.steamId && library.provenance.channel !== 'local-scan' && AppSettings.get('autoLoadProfile')) {
```
EricWehrly: why do we need the strong differentiation with local-scan? Could it leverage more of the existing flow?

client/src/steam/SteamApiClient.ts
```diff
@@ -81,14 +83,13 @@ export class SteamApiClient {
...
-        // seeded cache (e.g. the anonymous store's demo game list) await appDetailsCacheReady.
-        this.appDetailsCacheReady = this.appDetailsCache.init()
-            .then(() => this.bakedCacheLoader.seedIfNeeded())
+        this.bakedCacheLoader = new BakedCacheLoader()
+
+        // Seed AppDetailsCache from the baked release bundle. Fire-and-forget for scene startup
+        // in general (never blocks); callers that need the seeded cache (e.g. the anonymous
+        // store's demo game list) await appDetailsCacheReady. No readiness event needed here -
+        // AppDetailsCache.mergeMany makes any other writer racing this seed safe by construction.
+        this.appDetailsCacheReady = this.bakedCacheLoader.seedIfNeeded()
```
EricWehrly, line 92: Don't involve steamapiclient where it shouldn't be.
This also overloads its responsibilities, but that's a different discussion.

We have two signals we're tracking so that SteamIntegration can load demo games, 
This really feels like overloads across the board:
SteamIntegration is _also_ acting as a library manager where that feels a little beyond its responsibility AND we've just been drafting classes to manage libraries
We have two signals that steamapiclient is driving that could be two events, but really are one signal: Have we finished loading any pre-sseded cached data? We should only care about the pre-seeded appdata in this realm, and the pre-seeded artwork in the renderer realm, but ... 
that probably makes this an even bigger refactor. We should consider what makes the most sense.

